PR #11, opened 2019-05-30
Implementing Dependency injection

iruzevic:
### Added

- Added dependency injection in service locator inside class-main.php
- Changed Manifest.php method to fetch items to non static method.

### Changed

- Updating Manifest.php with new methods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a 
 
 _No documentation available about unreleased changes as of yet._
 
+## [0.7.0]
+
+### Added
+
+- Added dependency injection in service locator inside class-main.php
+- Changed Manifest.php method to fetch items to non static method.
+
+## [0.6.0]
+
+### Changed
+
+- Updating Manifest.php with new methods
+
 ## [0.5.0]
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ _No documentation available about unreleased changes as of yet._
 
 ### Added
 
-- Added dependency injection in service locator inside class-main.php
-- Changed Manifest.php method to fetch items to non static method.
+- Added DI instead of SL inside the class-main.php.
+- Changed methods used for fetching the manifest items inside Manifest.php to non static methods.
+
 
 ## [0.6.0]
 
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "source": "https://github.com/infinum/eightshift-libs"
   },
   "require": {
-    "php": ">=7.0",
+    "php": ">=7.2",
     "php-di/php-di": "^6.0"
   },
   "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "source": "https://github.com/infinum/eightshift-libs"
   },
   "require": {
-    "php": ">=7.0"
+    "php": ">=7.0",
+    "php-di/php-di": "^6.0"
   },
   "require-dev": {
     "infinum/coding-standards-wp": "^0.4.1",
--- a/src/assets/class-manifest.php
+++ b/src/assets/class-manifest.php
@@ -4,7 +4,7 @@
  *
  * It is used to provide manifest.json file location used with Webpack to fetch correct file locations.
  *
- * @since   0.6.0 Adding multiple methodes for easier extending.
+ * @since   0.6.0 Added multiple methods for easier extending.
  * @since   0.1.0
  * @package Eightshift_Libs\Assets
  */
@@ -12,6 +12,7 @@
 namespace Eightshift_Libs\Assets;
 
 use Eightshift_Libs\Core\Service;
+use Eightshift_Libs\Exception\Missing_Manifest;
 
 /**
  * Abstract class Manifest class.
@@ -50,7 +51,7 @@ abstract class Manifest implements Service, Manifest_Data {
    * @since 0.6.0 Init.
    */
   public function register_global_variable() : void {
-    \define( $this->get_global_variable_name(), $this->get_raw_data() );
+    define( $this->get_global_variable_name(), $this->get_raw_data() );
   }
 
   /**
@@ -58,9 +59,12 @@ abstract class Manifest implements Service, Manifest_Data {
    * This is used for cache busting assets.
    *
    * @param string $key File name key you want to get from manifest.
+   *
+   * @throws Exception\Missing_Manifest Throws error if manifest key is missing.
+   *
    * @return string Full path to asset.
    *
-   * @since 0.7.0 Changed to non static method.
+   * @since 0.7.0 Changed to non static method and added Exception for missing key.
    * @since 0.6.0 Init
    */
   public function get_assets_manifest_item( string $key ) : string {
@@ -70,13 +74,15 @@ abstract class Manifest implements Service, Manifest_Data {
 
     $data = $this->get_decoded_manifest_data();
 
-    $asset = $data->$key ?? '';
-
-    if ( empty( $asset ) ) {
-      return '';
+    if ( ! isset( $data[ $key ] ) ) {
+      $error_message = sprintf(
+        esc_html__( '%s is missing in manifest.json. Please check if provided key is correct.', 'eightshift-libs' ),
+        $key
+      );
+      throw Missing_Manifest::message( $error_message );
     }
 
-    return $this->get_assets_manifest_output_prefix() . $asset;
+    return $this->get_assets_manifest_output_prefix() . $data[ $key ];
   }
 
   /**
@@ -99,6 +105,7 @@ abstract class Manifest implements Service, Manifest_Data {
    *
    * @return string
    *
+   * @since 0.7.0 Fixed Exception msg.
    * @since 0.1.0
    */
   protected function get_raw_data() : string {
@@ -107,7 +114,7 @@ abstract class Manifest implements Service, Manifest_Data {
 
     if ( ! file_exists( $manifest ) ) {
       $error_message = esc_html__( 'manifest.json is missing. Bundle the theme before using it.', 'eightshift-libs' );
-      throw Exception\Missing_Manifest::message( $error_message );
+      throw Missing_Manifest::message( $error_message );
     }
 
     return implode( ' ', file( $manifest ) );
@@ -135,7 +142,7 @@ abstract class Manifest implements Service, Manifest_Data {
    * @since 0.6.0 Init
    */
   protected function get_decoded_manifest_data() {
-    $data = \json_decode( constant( $this->get_global_variable_name() ), true );
+    $data = json_decode( constant( $this->get_global_variable_name() ), true );
 
     if ( ! $data ) {
       return null;
@@ -153,6 +160,6 @@ abstract class Manifest implements Service, Manifest_Data {
    * @since 0.6.0
    */
   protected function get_assets_manifest_output_prefix() : string {
-    return \home_url( '/' );
+    return home_url( '/' );
   }
 }
--- a/src/assets/class-manifest.php
+++ b/src/assets/class-manifest.php
@@ -22,7 +22,7 @@ use Eightshift_Libs\Core\Service;
 abstract class Manifest implements Service, Manifest_Data {
 
   /**
-   * Global variable name contant.
+   * Global variable name constant.
    *
    * @var string
    *
@@ -43,7 +43,7 @@ abstract class Manifest implements Service, Manifest_Data {
 
   /**
    * Define global variable for assets manifest.
-   * Used to cache data insinde global variable so you don't fetch manifest.json file on every call.
+   * Used to cache data inside a global variable so you don't have to fetch manifest.json file on every call.
    *
    * @return void
    *
@@ -54,7 +54,7 @@ abstract class Manifest implements Service, Manifest_Data {
   }
 
   /**
-   * Return full path for specific asset from manifest.json
+   * Return full path for specific asset from manifest.json.
    * This is used for cache busting assets.
    *
    * @param string $key File name key you want to get from manifest.
@@ -78,7 +78,7 @@ abstract class Manifest implements Service, Manifest_Data {
   }
 
   /**
-   * Set global variable name to store cached data into.
+   * Get global variable name to store the cached data into.
    *
    * @return string
    *
@@ -111,7 +111,7 @@ abstract class Manifest implements Service, Manifest_Data {
   }
 
   /**
-   * Provide manifest.json url location.
+   * Get manifest.json url location.
    * If you are using a plugin or a different manifest location provide location with this method.
    *
    * @return string
@@ -141,7 +141,7 @@ abstract class Manifest implements Service, Manifest_Data {
   }
 
   /**
-   * Retrun string as a assets output prefix.
+   * Retrun string as an assets output prefix.
    * Override this if you are using lib for a plugin.
    *
    * @return string
--- a/src/assets/class-manifest.php
+++ b/src/assets/class-manifest.php
@@ -63,7 +63,7 @@ abstract class Manifest implements Service, Manifest_Data {
    * @since 0.7.0 Changed to non static method.
    * @since 0.6.0 Init
    */
-  public function get_assets_manifest_item( $key = null ) : string {
+  public function get_assets_manifest_item( string $key ) : string {
     if ( ! $key ) {
       return '';
     }
--- a/src/assets/class-manifest.php
+++ b/src/assets/class-manifest.php
@@ -71,9 +71,11 @@ abstract class Manifest implements Service, Manifest_Data {
     $data = $this->get_decoded_manifest_data();
 
     $asset = $data->$key ?? '';
+
     if ( empty( $asset ) ) {
       return '';
     }
+
     return $this->get_assets_manifest_output_prefix() . $asset;
   }
 
@@ -102,6 +104,7 @@ abstract class Manifest implements Service, Manifest_Data {
   protected function get_raw_data() : string {
 
     $manifest = $this->get_manifest_url();
+
     if ( ! file_exists( $manifest ) ) {
       $error_message = esc_html__( 'manifest.json is missing. Bundle the theme before using it.', 'eightshift-libs' );
       throw Exception\Missing_Manifest::message( $error_message );
@@ -116,7 +119,7 @@ abstract class Manifest implements Service, Manifest_Data {
    *
    * @return string
    *
-   * @since 0.6.0 Changed from abstract to predefined.
+   * @since 0.6.0 Changed from abstract method to prefilled.
    * @since 0.1.0
    */
   protected function get_manifest_url() : string {
@@ -132,7 +135,8 @@ abstract class Manifest implements Service, Manifest_Data {
    * @since 0.6.0 Init
    */
   protected function get_decoded_manifest_data() {
-    $data = \json_decode( constant( $this->get_global_variable_name() ) );
+    $data = \json_decode( constant( $this->get_global_variable_name() ), true );
+
     if ( ! $data ) {
       return null;
     }
--- a/src/assets/class-manifest.php
+++ b/src/assets/class-manifest.php
@@ -15,8 +15,11 @@ use Eightshift_Libs\Core\Service;
 
 /**
  * Abstract class Manifest class.
+ *
+ * @since 0.7.0 Added Manifest_Data Interface.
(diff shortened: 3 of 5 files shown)


src/assets/class-manifest.php
```diff
@@ -4,26 +4,90 @@
...
+ * @since 0.7.0 Added Manifest_Data Interface.
+ * @since 0.1.0 Init.
  */
-abstract class Manifest {
+abstract class Manifest implements Service, Manifest_Data {
 
   /**
-   * Provide manifest.json url location.
-   * You project must provide location for the manifest.json for this to work.
+   * Global variable name contant.
```
dingo-d:
```suggestion
   * Global variable name constant.
```

src/assets/class-manifest.php
```diff
@@ -4,26 +4,90 @@
...
+   *
+   * @since 0.6.0 Init.
+   */
+  public function register() : void {
+    add_action( 'init', [ $this, 'register_global_variable' ] );
+  }
+
+  /**
+   * Define global variable for assets manifest.
+   * Used to cache data insinde global variable so you don't fetch manifest.json file on every call.
```
dingo-d:
```suggestion
   * Used to cache data inside a global variable so you don't have to fetch manifest.json file on every call.
```

src/assets/class-manifest.php
```diff
@@ -4,26 +4,90 @@
...
+   * @return void
+   *
+   * @since 0.6.0 Init.
+   */
+  public function register_global_variable() : void {
+    \define( $this->get_global_variable_name(), $this->get_raw_data() );
+  }
+
+  /**
+   * Return full path for specific asset from manifest.json
```
dingo-d:
```suggestion
   * Return full path for specific asset from manifest.json.
```

src/assets/class-manifest.php
```diff
@@ -4,26 +4,90 @@
...
+   */
+  public function get_assets_manifest_item( string $key ) : string {
+    if ( ! $key ) {
+      return '';
+    }
+
+    $data = $this->get_decoded_manifest_data();
+
+    $asset = $data->$key ?? '';
+    if ( empty( $asset ) ) {
```
dingo-d: I'd add space before the conditional and after the conditional. Makes it a bit more readable.

src/assets/class-manifest.php
```diff
@@ -4,26 +4,90 @@
...
+      return '';
+    }
+    return $this->get_assets_manifest_output_prefix() . $asset;
+  }
+
+  /**
+   * Set global variable name to store cached data into.
```
dingo-d: Getter, not setter 🙂

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
    *
    * @since 0.1.0
    */
-  public function register_assets_manifest_data() : string {
+  protected function get_raw_data() : string {
 
     $manifest = $this->get_manifest_url();
     if ( ! file_exists( $manifest ) ) {
-      $error_message = esc_html__( 'manifest.json is missing. Bundle the theme before using it.', 'developer-portal' );
+      $error_message = esc_html__( 'manifest.json is missing. Bundle the theme before using it.', 'eightshift-libs' );
```
dingo-d: Add empty line before the conditional please 🙂

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
...
       throw Exception\Missing_Manifest::message( $error_message );
     }
 
     return implode( ' ', file( $manifest ) );
   }
+
+  /**
+   * Provide manifest.json url location.
```
dingo-d:
```suggestion
   * Get manifest.json url location.
```

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
...
+   * If you are using a plugin or a different manifest location provide location with this method.
+   *
+   * @return string
+   *
+   * @since 0.6.0 Changed from abstract to predefined.
```
dingo-d: Predefined what? This since is a bit unclear.

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
...
+  /**
+   * Return json_decoded manifest data, now you can call items by object key to get the value.
+   *
+   * @return object Manifest Object.
+   *
+   * @since 0.7.0 Changed to non static method.
+   * @since 0.6.0 Init
+   */
+  protected function get_decoded_manifest_data() {
+    $data = \json_decode( constant( $this->get_global_variable_name() ) );
```
dingo-d: This is neat, didn't use `constant()` before 🙂 👍 

Just one thing, add `true` after the first argument of `json_decode()` function.

Also, feel free to remove all `\` prefixes. This is a php function. Plus I've removed all those from my code, and everything works fine (if function is not found in the current namespace, php will try to find it in namespaces above and finally in the global namespace).

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
...
+  /**
+   * Retrun string as a assets output prefix.
+   * Override this if you are using lib for a plugin.
+   *
+   * @return string
+   *
+   * @since 0.6.0
+   */
+  protected function get_assets_manifest_output_prefix() : string {
+    return \home_url( '/' );
```
dingo-d: The name of the method and what it returns is totally confusing to me. Why is here `home_url()` function returned? Also `\` is not needed.

src/assets/class-manifest.php
```diff
@@ -35,14 +99,56 @@ abstract class Manifest {
...
+    $data = \json_decode( constant( $this->get_global_variable_name() ) );
+    if ( ! $data ) {
+      return null;
+    }
+
+    return $data;
+  }
+
+  /**
+   * Retrun string as a assets output prefix.
```
dingo-d:
```suggestion
   * Retrun string as an assets output prefix.
```

Can you also give an example of how this looks like in the end, in the comment? I'm not 100% sure what this does and why 🤷‍♂

src/assets/interface-manifest-data.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Interface for Assets.
```
dingo-d:
```suggestion
 * Assets manifest data interface.
 * 
 * Used to define the way manifest item is retrieved from the manifest file.
```

Or something along this line

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+    return array_map(
+      function( $class ) use ( $container ) {
+        return $container->get( $class );
+      },
+      array_keys( $services )
+    );
+  }
+
+  /**
+   * Get services classes array ad prepare it.
```
dingo-d:
```suggestion
   * Get services classes array and prepare it for dependency injection.
```

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+   * Key should be class name, and value empty array or DI items.
```
dingo-d:
```suggestion
   * Key should be a class name, and value should be an empty array or the dependencies of the class.
```

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+   *
+   * @return array
+   *
+   * @since 0.7.0 Init.
+   */
+  private function get_service_classes_prepared_array() : array {
+    $output = [];
+    foreach ( $this->get_service_classes() as $key => $value ) {
```
dingo-d:
```suggestion
    foreach ( $this->get_service_classes() as $class => $dependencies ) {
```

Makes it more clear what you are looping through.

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+      if ( gettype( $value ) !== 'array' ) {
+        $output[ $value ] = [];
```
dingo-d: Should this be `$output[ $key ]` (also `$key` == `$class`)?

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+        $output[ $key ] = $value;
       }
+    }
+
+    return $output;
+  }
+
+  /**
+   * Implement PHP-DI.
+   * Create Object used for PHP-DI to create Dependency Injection methods.
```
dingo-d:
```suggestion
   * Build and return a DI container.
   * Wire all the dependencies automatically, based on the provided array of class => dependencies from the get_di_items().
```

Something like this.

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+   *
+   * @param array $services Array of service.
+   * @return object
+   *
+   * @since 0.7.0 Init.
+   */
+  private function get_di_container( array $services ) {
+    $builder = new ContainerBuilder();
+    $builder->useAnnotations( false );
+    $builder->useAutowiring( true );
```
dingo-d: Actually I think we can safely remove both `$builder->useAnnotations( false );` and `$builder->useAutowiring( true );` as those are defaults. I think I added them out of paranoia 🤣

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+   * If you pass a class use PHP-DI to prepare if not just output it.
+   *
+   * @param array $items Array of classes/parameters to push in constructor.
+   * @return array
+   *
+   * @since 0.7.0 Init
+   */
+  private function get_di_items( array $items ) : array {
+    return array_map(
+      function( $class ) {
```
dingo-d: This doesn't have to be a class, no? Maybe `$dependency` would be better? 🙂

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+  /**
+   * Return prepared Dependency Injection objects.
+   * If you pass a class use PHP-DI to prepare if not just output it.
+   *
+   * @param array $items Array of classes/parameters to push in constructor.
+   * @return array
+   *
+   * @since 0.7.0 Init
+   */
+  private function get_di_items( array $items ) : array {
```
dingo-d:
```suggestion
  private function get_di_dependencies( array $dependencies ) : array {
```
Semantically makes more sense to me

src/class-main.php
```diff
@@ -77,17 +81,105 @@ public function register_services() : void {
...
+    return array_map(
+      function( $class ) {
+        if ( class_exists( $class ) ) {
+          return \DI\get( $class );
+        }
+        return $class;
+      },
+      $items
```
dingo-d: If the [above comment](https://github.com/infinum/eightshift-libs/pull/11/files#r289329512) is accepted, don't forget to change this to `$dependencies` as well 🙂

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: Manifest gives you and output like this:
```
{
  "application.css": "wp-content/themes/wp-boilerplate/skin/public/application-9d11e4277e8ca74d10ed.css",
  "application.js": "wp-content/themes/wp-boilerplate/skin/public/application.js",
  "applicationAdmin.css": "wp-content/themes/wp-boilerplate/skin/public/applicationAdmin-9d11e4277e8ca74d10ed.css",
  "applicationAdmin.js": "wp-content/themes/wp-boilerplate/skin/public/applicationAdmin.js",
  "jquery.min.js": "wp-content/themes/wp-boilerplate/skin/public/jquery.min.js",
  "logo.svg": "wp-content/themes/wp-boilerplate/skin/public/logo.svg"
}
```

this method adds site url to the final output url.
I have added home_url as a default (using it in theme), if you are using it in plugin you will override this method

src/assets/class-manifest.php
iruzevic: example in previous comment

src/class-main.php
iruzevic: 👍

src/class-main.php
iruzevic: this is because you can create something like this in you setup:
```
protected function get_service_classes() : array {
    return [
      Assets\Manifest::class,
      Admin\Admin::class => [ Assets\Manifest::class ],
      Admin\Login::class,
      Admin\Media::class,
      Admin\Widgets::class,
      Menu\Menu::class,
      Theme\Theme::class => [ Assets\Manifest::class ],
    ];
  }```

so this array is mixed.
I have looped this array and created it the correct way, so it is easier in the later setup to fetch dependencies.

File: src/class-main.php (hunk 18 above)
Comment (iruzevic) [reply src/class-main.php]: 👍

File: src/class-main.php (hunk 20 above)
Comment (iruzevic) [reply src/class-main.php]: 👍

File: src/class-main.php (hunk 19 above)
Comment (iruzevic) [reply src/class-main.php]: 👍

File: src/class-main.php (hunk 21 above)
Comment (iruzevic) [reply src/class-main.php]: 👍

File: CHANGELOG.md [hunk 22]
```diff
@@ -7,6 +7,19 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
 
 _No documentation available about unreleased changes as of yet._
 
+## [0.7.0]
+
+### Added
+
+- Added dependency injection in service locator inside class-main.php
```
dingo-d: I'd rephrase this. DI and SL are two ways of implementing IoC container, maybe say something like

```
Added DI instead of SL inside the class-main.php
```

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
...
+- Changed Manifest.php method to fetch items to non static method.
```
dingo-d: Same here

```suggestion
- Changed methods used for fetching the manifest items inside Manifest.php to non static methods.
```

composer.json
```diff
@@ -19,7 +19,8 @@
     "source": "https://github.com/infinum/eightshift-libs"
   },
   "require": {
-    "php": ">=7.0"
+    "php": ">=7.0",
```
dingo-d: Should we up this or leave it as is? 7.0 has reached end of life and is no longer maintained.

https://www.php.net/supported-versions.php

src/assets/class-manifest.php
```diff
@@ -4,26 +4,92 @@
  *
  * It is used to provide manifest.json file location used with Webpack to fetch correct file locations.
  *
+ * @since   0.6.0 Adding multiple methodes for easier extending.
```
dingo-d:
```suggestion
 * @since   0.6.0 Added multiple methods for easier extending.
```

src/assets/class-manifest.php
```diff
@@ -4,26 +4,92 @@
...
+  /**
+   * Define global variable for assets manifest.
+   * Used to cache data inside a global variable so you don't have to fetch manifest.json file on every call.
+   *
+   * @return void
+   *
+   * @since 0.6.0 Init.
+   */
+  public function register_global_variable() : void {
+    \define( $this->get_global_variable_name(), $this->get_raw_data() );
```
dingo-d:
```suggestion
    define( $this->get_global_variable_name(), $this->get_raw_data() );
```

src/assets/class-manifest.php
```diff
@@ -35,14 +101,58 @@ abstract class Manifest {
...
+   * Get manifest.json url location.
+   * If you are using a plugin or a different manifest location provide location with this method.
+   *
+   * @return string
+   *
+   * @since 0.6.0 Changed from abstract method to prefilled.
+   * @since 0.1.0
+   */
+  protected function get_manifest_url() : string {
+    return get_template_directory() . '/skin/public/manifest.json';
```
dingo-d: Are we planning on using the libs in themes only? If not, then we may need to figure out how to handle this smarter to cover the plugin case.

For my view loading in the plugin I'm writing I have something like

```php
  /**
   * Get the possible locations for the view.
   *
   * @param string $uri URI of the view to get the locations for.
   *
   * @return array Array of possible locations.
   */
  protected function get_locations( $uri ) : array {
    return [
      trailingslashit( \get_stylesheet_directory() ) . $uri,
      trailingslashit( \get_template_directory() ) . $uri,
      trailingslashit( dirname( __DIR__, 2 ) ) . $uri,
    ];
  }```

Then, based on the provided path I check if it's readable (in the method used for validating the uri)

```php
foreach ( $this->get_locations( $uri ) as $location ) {
      if ( is_readable( $location ) ) {
        return $location;
      }
    }
```

Something like this could be done to cover all the cases (themes and plugins.)

src/assets/class-manifest.php
```diff
@@ -35,14 +101,58 @@ abstract class Manifest {
...
+  /**
+   * Return json_decoded manifest data, now you can call items by object key to get the value.
+   *
+   * @return object Manifest Object.
+   *
+   * @since 0.7.0 Changed to non static method.
+   * @since 0.6.0 Init
+   */
+  protected function get_decoded_manifest_data() {
+    $data = \json_decode( constant( $this->get_global_variable_name() ), true );
```
dingo-d:
```suggestion
    $data = json_decode( constant( $this->get_global_variable_name() ), true );
```

composer.json
IvanGrginovInf: If there's a single `public function someMethod() : void {`, this needs to be `>=7.2`

src/class-main.php
```diff
@@ -77,17 +81,104 @@ public function register_services() : void {
...
-    $this->services = array_map(
-      [ $this, 'instantiate_service' ],
-      $classes
-    );
     array_walk(
       $this->services,
-      function( Service $service ) {
-        $service->register();
+      function( $class ) {
+        if ( ! $class instanceof Service ) {
```
dingo-d: I think I changed this to `Registerable`, because I had one class that just implemented `Registerable` and not `Service`. Just something to think about.

src/class-main.php
```diff
@@ -77,17 +81,104 @@ public function register_services() : void {
...
+   * @param array $services Array of service.
+   * @return object
+   *
+   * @since 0.7.0 Init.
+   */
+  private function get_di_container( array $services ) {
+    $builder = new ContainerBuilder();
+
+    $definitions = [];
+    foreach ( $services as  $key => $values ) {
```
dingo-d: I'd change `$key` and `$value` into something more descriptive. Also there is an extra space in front of `$key`.

CHANGELOG.md
iruzevic: 👍

CHANGELOG.md
iruzevic: 👍

composer.json
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: 👍

src/class-main.php
iruzevic: 👍

src/assets/class-manifest.php
iruzevic: good idea but I would like to think more about this so I will leave this out for now and add an issue for the future

iruzevic:
all corrections are added. Can you please approve so I can start on the Gutenberg blocks update

src/assets/class-manifest.php
```diff
@@ -4,26 +4,98 @@
...
+    }
+
+    $data = $this->get_decoded_manifest_data();
+
+    if ( ! isset( $data[ $key ] ) ) {
+      $error_message = sprintf(
+        esc_html__( '%s is missing in manifest.json. Please check if provided key is correct.', 'eightshift-libs' ),
+        $key
+      );
+      throw Missing_Manifest::message( $error_message );
```
dingo-d: In case manifest is missing you'd get a type error first, because you'd throw an error, but the expected return type is a `string`. So you can remove the return type hint to fix this (that's what I did whenever I had cases like this).

PR #13, opened 2019-07-08
Eightshift Libs

iruzevic:


src/assets/class-manifest.php
```diff
@@ -85,6 +86,25 @@ public function get_assets_manifest_item( string $key ) : string {
     return $this->get_assets_manifest_output_prefix() . $data[ $key ];
   }
 
+  /**
+   * Return json_decoded manifest data, now you can call items by object key to get the value.
+   *
+   * @return object Manifest Object.
```
dingo-d: If `$data` is a falsy value you return `null`, which is not an object. Another thing is that this is wrong return doctype, as the return value is an array (you've set the second parameter of `json_decode` to `true`).

So I'd change the return type to `array` (in the docs and as a return type hint, as that is available in PHP 7.0), and in the `if( ! $data )` conditional I'd return empty array (a falsy value).

iruzevic: Fixed tnx

PR #18, opened 2019-09-01
[WIP] Automated tests

dingo-d:
⚠️**This PR is a work in progress, do not merge** ⚠️

This PR sets up the automated test suite for running unit and integration tests for the library with WordPress.

- [x] Set up the phpunit
- [x] Set up code coverage driver (pcov)
- [x] Write unit tests for main functionality
- [x] Write unit tests for exceptions
- [x] Write unit tests for custom post types
- [x] Write unit tests for custom taxonomies
- [x] Write unit tests for asset manifest
- [ ] Write integration tests with WordPress


dingo-d:
I've written all the unit tests.

I've removed the unused `instantiate_service` method, seems not to be used anywhere 🤷‍♂ 

Also, there is a part

https://github.com/infinum/eightshift-libs/blob/28c4b2fa26bdbe3215df59cc8dfdee145b8c4929/src/class-main.php#L80-L83

That doesn't work on its own.

If I'm correct, this usage is only useful when a factory pattern is used with a static variable and when you want to define a shared instance of the object.

On its own, I don't think it has much use, but it doesn't hurt to have it in the code.

Current code coverage (unit tests):

libs-tests

I need to finish the integration tests.

The to do for that is:

Move examples to the 'fake plugin' inside test fixtures. Add it as a plugin to test the integration test in, and see if the plugin actual functionalities work within WordPress. This means, testing to see if the custom taxonomies and custom post types are registered if the REST endpoints are registered etc.

It would be a great idea to have this be a fully functioning plugin (with abstract classes etc.) so that the tests have more impact. 
Any help in that regard will be welcomed @infinum/eightshift-team 🙂

src/class-main.php
```diff
@@ -68,11 +68,11 @@ public function get_register_action_hook() : string {
...
    * Register the individual services with optional dependency injection.
    *
-   * @throws Exception\Invalid_Service If a service is not valid.
-   *
    * @return void
    *
-   * @since 0.7.0 Dependency Injection Refactoring
+   * @throws \Exception
+   *
+   * @throws Exception\Invalid_Service If a service is not valid.*@since 0.7.0 Dependency Injection Refactoring
```
IvanGrginovInf: Can you please fix the formatting here?

src/class-main.php
```diff
@@ -101,6 +101,7 @@ function( $class ) {
    *
    * @return array
    *
+   * @throws \Exception
```
IvanGrginovInf: I can't seem to find any throwing in this method. If it does throw for some other reason can you please add a reason why in the doc comment?

src/class-main.php
```diff
@@ -109,7 +110,7 @@ private function get_service_classes_with_di() : array {
     $container = $this->get_di_container( $services );
 
     return array_map(
-      function( $class ) use ( $container ) {
+      static function($class ) use ( $container ) {
```
IvanGrginovInf: There should be a space before `$class` here: `function($class )`

src/class-main.php
```diff
@@ -128,7 +129,7 @@ private function get_service_classes_prepared_array() : array {
     $output = [];
 
     foreach ( $this->get_service_classes() as $class => $dependencies ) {
-      if ( gettype( $dependencies ) !== 'array' ) {
+      if (!is_array($dependencies)) {
```
IvanGrginovInf: ⬆️ also you should probably go through this with code sniffer for formatting.

src/class-main.php
```diff
@@ -146,6 +147,7 @@ private function get_service_classes_prepared_array() : array {
    * @param array $services Array of service.
    * @return object
    *
+   * @throws \Exception
```
IvanGrginovInf: Same as above, it doesn't seem to explicitly throw. Can you please add an explanation why / when it throws.

dingo-d:
Closed because we'll set up codeception on libs

Issue #21, opened 2019-10-29
Make routes interfaces less granular


iruzevic:
I kinda feel like this and the above interfaces are almost too granular. There is Callable_Field, Callable_Route, Registrable_Field, Registrable_Route and Route_Security.

Do you really need all of them?

In the plugin I made I ended up with just one Route interface with method definitions and Rest_Callable interface for the rest_callback.
Then I added a Base_Route abstract class for creating routes and left every class in charge of defining the implementation of the rest_callback and the get_callback_arguments method which may or may not include permission_callback because not all routes need one.

I'm not sure how the implementation would look in this case.

iruzevic:
on pull request https://github.com/infinum/eightshift-libs/pull/22

iruzevic:
this is done

PR #22, opened 2019-10-30
Feature/libs blocks merge

iruzevic:
- fixing caching issues on blocks
- Blocks reordering and renaming method names
- Adding New config methods for rest
- Refactoring Rest routes and fields methods (https://github.com/infinum/eightshift-libs/issues/21)


src/interface-config-data.php
```diff
@@ -59,6 +59,24 @@ public static function get_project_prefix() : string;
...
   public static function get_project_env() : string;
 
+  /**
+   * Method that returns project REST-API namespace.
+   *
+   * Used for namespacing projects REST-API routes and fields.
+   *
+   * @since 2.0.0 Added in the project
+   */
+  public static function get_project_routes_namespace() : string;
```
dingo-d: Why static?

iruzevic: Yes, but unfortunately this is the only way to cover all the aspects for the code (internal code - classes, templates and WP templates). Only this config file is made of static methods and nothing else and in the classes we are using DI abut in the WP templates we are using static methods. I think this is ok, we have interface for the config file so it can simply be mocked.

PR #29, opened 2019-11-14
adding gitattributes and di container cache

iruzevic:


README.md
```diff
@@ -30,6 +30,9 @@ We provide some helpers, abstract classes, interfaces and abstractions on origin
 
 For detail documentation please check on [Eightshift Boilerplate Wiki](https://github.com/infinum/eightshift-boilerplate/wiki)
 
+## Requirement
+To run PHP-DI with caching system enabled your server must have [APC](https://www.php.net/manual/en/book.apc.php) extension instaled.
```
dingo-d: If I'm not mistaken APC was replaced with OPcache (after PHP5.5), so is this really necessary?

src/class-main.php
```diff
@@ -143,6 +155,10 @@ private function get_service_classes_prepared_array() : array {
   private function get_di_container( array $services ) {
     $builder = new ContainerBuilder();
 
+    if ( $this->get_env() === 'production' ) {
+      $builder->enableDefinitionCache();
```
dingo-d: I usually use `$builder->enableCompilation( __DIR__ );` didn't see some negative performance so far.

Would love to compare the two and profile an application to see the results.

iruzevic: I have used this method in the #Cache section.

http://php-di.org/doc/performances.html#caching

Not sure if this is the bes option to use. Any thoughts?

README.md
iruzevic: Yep in the documentation it is stated that you need APC cache. I have opcache on my local env and it throw fatal error. I have set the APC extension and it worked

src/class-main.php
IvanGrginovInf: >If you make heavy use of those features, and if it slows down your application you can enable the caching system
http://php-di.org/doc/performances.html#caching

From the wording it would suggest we don't really need it unless things get slow?

PR #32, opened 2020-01-03
Update enqueue functionality

dingo-d:
I've added an abstract class to avoid repeating the method calls for dependencies. Also, config interface has static methods so the handle names and other config related methods had to be changed into static calls.

This PR should enable people to add additional handlers in their enqueue classes with something like:

```php
use Eightshift_Libs\Enqueue\Enqueue_Admin;

class My_Enqeueue extends Enqueue_Admin {
  public function enqueue_scripts() {
    parent::enqueue_scripts();

    // Enqueue additional scripts?
  }

  protected function get_admin_script_dependencies() {
    return [ 'wp-pointer', 'jquery-ui-sortable' ];
  }

  protected function get_localizations() {
    return [
      'randomStuff' => esc_html__( 'Hi!', 'text-domain' );
    ];
  }
}
```

Feel free to test before approving this 😄

### DIFF
--- a/src/enqueue/class-enqueue-admin.php
+++ b/src/enqueue/class-enqueue-admin.php
@@ -21,6 +21,12 @@ use Eightshift_Libs\Core\Config_Data;
  * @since 2.0.0
  */
 class Enqueue_Admin extends Assets {
+
+  const IN_FOOTER = true;
+
+  const ADMIN_SCRIPT_URI = 'applicationAdmin.js';
+  const ADMIN_STYLE_URI  = 'applicationAdmin.css';
+
   /**
    * Instance variable of project config data.
    *
@@ -80,7 +86,7 @@ class Enqueue_Admin extends Assets {
 
     \wp_register_style(
       $handle,
-      $this->manifest->get_assets_manifest_item( 'applicationAdmin.css' ),
+      $this->manifest->get_assets_manifest_item( static::ADMIN_STYLE_URI ),
       $this->get_admin_style_dependencies(),
       $this->config::get_project_version(),
       $this->get_media()
@@ -104,10 +110,10 @@ class Enqueue_Admin extends Assets {
 
     \wp_register_script(
       $handle,
-      $this->manifest->get_assets_manifest_item( 'applicationAdmin.js' ),
+      $this->manifest->get_assets_manifest_item( static::ADMIN_SCRIPT_URI ),
       $this->get_admin_script_dependencies(),
       $this->config::get_project_version(),
-      true
+      $this->script_in_footer()
     );
 
     \wp_enqueue_script( $handle );
--- a/src/enqueue/class-enqueue-admin.php
+++ b/src/enqueue/class-enqueue-admin.php
@@ -20,8 +20,7 @@ use Eightshift_Libs\Core\Config_Data;
  *
  * @since 2.0.0
  */
-class Enqueue_Admin implements Service {
-
+class Enqueue_Admin extends Assets {
   /**
    * Instance variable of project config data.
    *
@@ -73,17 +72,21 @@ class Enqueue_Admin implements Service {
    * @return void
    *
    * @since 2.0.0
+   * @since 2.0.3 Added methods for overrides.
+   *              Fixed static calls from config class.
    */
   public function enqueue_styles() {
-    $handler = "{$this->config->get_project_prefix()}-styles";
+    $handle = "{$this->config::get_project_prefix()}-styles";
 
     \wp_register_style(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'applicationAdmin.css' ),
-      [],
-      $this->config->get_project_version()
+      $this->get_admin_style_dependencies(),
+      $this->config::get_project_version(),
+      $this->get_media()
     );
-    \wp_enqueue_style( $handler );
+
+    \wp_enqueue_style( $handle );
 
   }
 
@@ -93,17 +96,24 @@ class Enqueue_Admin implements Service {
    * @return void
    *
    * @since 2.0.0
+   * @since 2.0.3 Added methods for overrides.
+   *              Fixed static calls from config class.
    */
   public function enqueue_scripts() {
-    $handler = "{$this->config->get_project_prefix()}-scripts";
+    $handle = "{$this->config::get_project_prefix()}-scripts";
 
     \wp_register_script(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'applicationAdmin.js' ),
-      [],
-      $this->config->get_project_version(),
+      $this->get_admin_script_dependencies(),
+      $this->config::get_project_version(),
       true
     );
-    \wp_enqueue_script( $handler );
+
+    \wp_enqueue_script( $handle );
+
+    foreach ( $this->get_localizations() as $object_name => $data_array ) {
+      \wp_localize_script( $handle, $object_name, $data_array );
+    }
   }
 }
--- a/src/enqueue/class-enqueue-blocks.php
+++ b/src/enqueue/class-enqueue-blocks.php
@@ -184,6 +184,13 @@ class Enqueue_Blocks extends Assets {
     return [ "{$this->config::get_project_prefix()}-block-style" ];
   }
 
+    /**
+     * List of admin script dependencies
+     *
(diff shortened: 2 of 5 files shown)


src/enqueue/class-enqueue-blocks.php
```diff
@@ -108,60 +98,97 @@ public function enqueue_block_editor_script() {
...
+    $handler = "{$this->config::get_project_prefix()}-block-style";
 
     \wp_register_style(
       $handler,
       $this->manifest->get_assets_manifest_item( 'applicationBlocks.css' ),
-      [],
-      $this->config->get_project_version(),
+      $this->get_frontend_style_dependencies(),
+      $this->config::get_project_version(),
       false
```
iruzevic: media override, could you please add it here also

src/enqueue/class-enqueue-admin.php
```diff
@@ -73,17 +72,21 @@ public function register() {
...
+   *              Fixed static calls from config class.
    */
   public function enqueue_styles() {
-    $handler = "{$this->config->get_project_prefix()}-styles";
+    $handle = "{$this->config::get_project_prefix()}-styles";
 
     \wp_register_style(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'applicationAdmin.css' ),
```
iruzevic: we should think about method to change the name of the file?

src/enqueue/class-enqueue-admin.php
```diff
@@ -93,17 +96,24 @@ public function enqueue_styles() {
...
+   *              Fixed static calls from config class.
    */
   public function enqueue_scripts() {
-    $handler = "{$this->config->get_project_prefix()}-scripts";
+    $handle = "{$this->config::get_project_prefix()}-scripts";
 
     \wp_register_script(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'applicationAdmin.js' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-blocks.php
```diff
@@ -76,29 +76,19 @@ public function register() {
...
    *
    * @return void
    */
   public function enqueue_block_editor_script() {
-    $handler = "{$this->config->get_project_prefix()}-block-editor-scripts";
+    $handler = "{$this->config::get_project_prefix()}-block-editor-scripts";
 
     \wp_register_script(
       $handler,
       $this->manifest->get_assets_manifest_item( 'applicationBlocksEditor.js' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-blocks.php
```diff
@@ -108,60 +98,97 @@ public function enqueue_block_editor_script() {
...
    * @return void
    */
   public function enqueue_block_editor_style() {
-    $handler    = "{$this->config->get_project_prefix()}-block-editor-style";
-    $dependency = "{$this->config->get_project_prefix()}-block-style";
+    $handler = "{$this->config::get_project_prefix()}-block-editor-style";
 
     \wp_register_style(
       $handler,
       $this->manifest->get_assets_manifest_item( 'applicationBlocksEditor.css' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-blocks.php
```diff
@@ -108,60 +98,97 @@ public function enqueue_block_editor_script() {
...
    *
    * @return void
    */
   public function enqueue_block_style() {
-    $handler = "{$this->config->get_project_prefix()}-block-style";
+    $handler = "{$this->config::get_project_prefix()}-block-style";
 
     \wp_register_style(
       $handler,
       $this->manifest->get_assets_manifest_item( 'applicationBlocks.css' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-blocks.php
```diff
@@ -108,60 +98,97 @@ public function enqueue_block_editor_script() {
...
    *
    * @return void
    */
   public function enqueue_block_script() {
-    $handler = "{$this->config->get_project_prefix()}-block-scripts";
+    $handler = "{$this->config::get_project_prefix()}-block-scripts";
 
     \wp_register_script(
       $handler,
       $this->manifest->get_assets_manifest_item( 'applicationBlocks.js' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-theme.php
```diff
@@ -64,50 +64,52 @@ public function register() {
...
+   *              Fixed static calls from config class.
    */
   public function enqueue_styles() {
-    $handler = "{$this->config->get_project_prefix()}-theme-styles";
+    $handle = "{$this->config::get_project_prefix()}-theme-styles";
 
     \wp_register_style(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'application.css' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-theme.php
```diff
@@ -64,50 +64,52 @@ public function register() {
...
 
     \wp_register_script(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'application.js' ),
-      [],
-      $this->config->get_project_version(),
+      $this->get_frontend_script_dependencies(),
+      $this->config::get_project_version(),
       true
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-theme.php
```diff
@@ -64,50 +64,52 @@ public function register() {
...
+   *              Fixed static calls from config class.
    */
   public function enqueue_scripts() {
-    $handler = "{$this->config->get_project_prefix()}-scripts";
+    $handle = "{$this->config::get_project_prefix()}-scripts";
 
     \wp_register_script(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'application.js' ),
```
iruzevic: just like 🆙

src/enqueue/class-enqueue-admin.php
```diff
@@ -93,17 +96,24 @@ public function enqueue_styles() {
...
-      $this->config->get_project_version(),
+      $this->get_admin_script_dependencies(),
+      $this->config::get_project_version(),
       true
     );
-    \wp_enqueue_script( $handler );
+
+    \wp_enqueue_script( $handle );
+
+    foreach ( $this->get_localizations() as $object_name => $data_array ) {
```
iruzevic: If there get_localizations is empty we can skip this loop. Can we add a conditional before it?

src/enqueue/class-enqueue-admin.php
```diff
@@ -93,17 +96,24 @@ public function enqueue_styles() {
...
 
     \wp_register_script(
-      $handler,
+      $handle,
       $this->manifest->get_assets_manifest_item( 'applicationAdmin.js' ),
-      [],
-      $this->config->get_project_version(),
+      $this->get_admin_script_dependencies(),
+      $this->config::get_project_version(),
       true
```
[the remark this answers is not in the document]
dingo-d: Thought about it, but then I thought that there was a good reason why you put it in the footer so I didn't change it 😄 I can add it :+1:

src/enqueue/class-enqueue-admin.php
dingo-d: Change the name of `applicationAdmin.css`? Isn't this webpack specific?

iruzevic: yes but you can change it, if you change it in webpack we need to change it here also.

src/enqueue/class-enqueue-admin.php
dingo-d: No need. By default, the `get_localizations()` method returns an empty array, and if you pass am empty array to `foreach` nothing will happen (no error no nothing). So it's ok to leave it here 🙂

iruzevic:
@dingo-d when do you think you will have this done?

dingo-d:
Updated

iruzevic:
Please add change log

dingo-d:
@IvanGrginovInf @kmaric69 @gabriel-glo Please check this PR 🙂

dingo-d:
I've updated the docblock @IvanGrginovInf

PR #40, opened 2020-02-17
Add components helper class.

IvanGrginovInf:
Changelog
---
* Added `class-components.php` helper for using components in WordPress views and other components

Part of https://github.com/infinum/eightshift-frontend-libs/pull/83

### DIFF
--- a/src/helpers/class-components.php
+++ b/src/helpers/class-components.php
@@ -30,7 +30,7 @@ class Components {
     } elseif ( is_string( $variable ) ) {
       $output = $variable;
     } else {
-      Component_Exception::not_string_or_variable( $variable );
+      Component_Exception::throw_not_string_or_variable( $variable );
     }
 
     return $output;
@@ -67,7 +67,7 @@ class Components {
     $template = \locate_template( $component_path );
 
     if ( empty( $template ) ) {
-      Component_Exception::unable_to_locate_component( $component_path );
+      Component_Exception::throw_unable_to_locate_component( $component_path );
     }
     ob_start();
     require $template;
--- a/src/helpers/class-components.php
+++ b/src/helpers/class-components.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Helpers for components
+ *
+ * @package Eightshift_Libs\Helpers
+ */
+
+declare( strict_types=1 );
+
+namespace Eightshift_Libs\Helpers;
+
+/**
+ * Helpers for components
+ *
+ * @since 4.0.0
+ */
+class Components {
+
+  /**
+   * Makes sure the output is string. Useful for converting an array of components into a string.
+   *
+   * @param  array|string $array_or_string Variable we need to convert into a string.
+   * @return string
+   */
+  public static function ensure_string( $array_or_string ) : string {
+    $output = '';
+
+    if ( is_array( $array_or_string ) ) {
+      $output = implode( '', $array_or_string );
+    } elseif ( is_string( $array_or_string ) ) {
+      $output = $array_or_string;
+    }
+
+    return $output;
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/helpers/class-components.php
```diff
@@ -0,0 +1,70 @@
...
+
+    // Detect if user passed component name or path.
+    if ( strpos( $component, '.php' ) !== false ) {
+      $template = \locate_template( $component );
+    } else {
+      $template = \locate_template( "src/blocks/components/$component/$component.php" );
+    }
+
+    if ( empty( $template ) ) {
+      throw new \Exception( "Unable to locate component: \"{$component}\"" );
```
iruzevic: write Exception using this method https://github.com/infinum/eightshift-libs/tree/develop/src/exception

src/helpers/class-components.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+    ob_start();
+    require $template;
+    return ob_get_clean();
```
iruzevic: do we need to unset variables here to remove possibility of pouting the components 

https://github.com/infinum/eightshift-libs/blob/develop/src/blocks/class-blocks.php#L340

src/helpers/class-components.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+class Components {
+
+  /**
+   * Makes sure the output is string. Useful for converting an array of components into a string.
+   *
+   * @param  array|string $array_or_string Variable we need to convert into a string.
+   * @return string
+   */
+  public static function ensure_string( $array_or_string ) : string {
```
dingo-d:
```suggestion
  public static function ensure_string( $variable ) : string {
```

Just reads weird when I see `$array_or_string`, there is already a docblock that specifies `array|string`. Don't think we need to be so verbose 😄

src/helpers/class-components.php
```diff
@@ -0,0 +1,70 @@
...
+    $output = '';
+
+    if ( is_array( $array_or_string ) ) {
+      $output = implode( '', $array_or_string );
+    } elseif ( is_string( $array_or_string ) ) {
+      $output = $array_or_string;
+    }
+
+    return $output;
+  }
```
dingo-d: What happens when you pass something that is not `array` or `string`? Like integer?

src/helpers/class-components.php
IvanGrginovInf: Makes sense although I'd imagine `unset()` would need to be called before the `include() / require()` otherwise it doesn't do anything.

And in this case **it's not needed** as the only 2 variables you have before requiring the template are the 2 you need.

src/helpers/class-components.php
IvanGrginovInf: Hmm, yeah didn't thought about that, I'll make it throw an exception in that case.

src/helpers/class-components.php
iruzevic: 👍

src/helpers/class-components.php
IvanGrginovInf: What can I say, I'm a sucker for verbosity. Will change. 👍

IvanGrginovInf:
@dingo-d & @iruzevic implemented changes.

src/helpers/class-components.php
dingo-d: Or you could have used `strval()`?

https://www.php.net/manual/en/function.strval.php

IvanGrginovInf: Sure I could cast it to string but I'd rather if it errored out with a helpful msg on invalid use rather than the function trying to make it work or silently failing.

It's the same reasoning why the `render()` method below does the same thing (our previous implementations were always "If the template doesn't exist, just do nothing, a.k.a silently fail"). Debugging that is hard.

PR #44, opened 2020-02-29
Add parent wrapper attribute.

IvanGrginovInf:
Changelog
---
* Added ability to wrap components with parent class on render

### DIFF
--- a/src/helpers/class-components.php
+++ b/src/helpers/class-components.php
@@ -48,6 +48,10 @@ class Components {
 
   /**
    * Renders a components and (optionally) passes some attributes to it.
+   * 
+   * Note about "parentClass" attribute: If provided, the component will be wrapped with a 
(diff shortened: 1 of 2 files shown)


src/helpers/class-components.php
```diff
@@ -59,18 +63,29 @@ public static function render( string $component, array $attributes = [] ) {
...
     } else {
-      $component_path = "src/blocks/components/$component/$component.php";
+      $template = locate_template( "src/blocks/components/$component/$component.php" );
     }
 
-    $template = \locate_template( $component_path );
-
     if ( empty( $template ) ) {
-      Component_Exception::throw_unable_to_locate_component( $component_path );
+      throw new \Exception( "Unable to locate component: \"{$component}\"" );
```
iruzevic: Why have you changed this method?

PR #45, opened 2020-03-02
Feature/helpers

iruzevic:
- Changing columns post type class name
- Moving shortcode class from general to helpers

### DIFF
--- a/src/helpers/class-shortcode.php
+++ b/src/helpers/class-shortcode.php
@@ -1,7 +1,6 @@
 <?php
 /**
  * The Shortcode specific functionality.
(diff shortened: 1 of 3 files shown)


src/helpers/class-shortcode.php
```diff
@@ -1,15 +1,14 @@
 <?php
 /**
  * The Shortcode specific functionality.
- * Used in admin or theme side.
  *
  * @since   1.0.0
  * @package Eightshift_Libs\General
```
IvanGrginovInf:
```suggestion
 * @package Eightshift_Libs\Helpers
```

PR #46, opened 2020-03-04
Fixing enqueue classes for missing method

iruzevic:
This build was seriously broken.

- There was a method missing in enqueue classes.


src/enqueue/class-assets.php
```diff
@@ -21,11 +21,17 @@
  */
 abstract class Assets implements Service {
 
+  /**
+   * Media style const
```
dingo-d: Don't think the `const` part is necessary here. I mean they are constants, kinda self-explanatory 😄

PR #55, opened 2020-05-14
adding responsive-selector-helper

iruzevic:
- Used in new colums block

### DIFF
--- a/src/helpers/class-components.php
+++ b/src/helpers/class-components.php
@@ -110,7 +110,7 @@ class Components {
     $output = [];
 
     foreach ( $items as $item_key => $item_value ) {
-      if ( empty( $item_value ) ) {
+      if ( empty( $item_value ) || $item_value === false ) {
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


src/helpers/class-components.php
```diff
@@ -90,4 +90,37 @@ public static function render( string $component, array $attributes = [] ) {
 
     return ob_get_clean();
   }
+
+  /**
+   * Create responsive selectors used for responsive attributes.
+   *
+   * @param array   $items        Array of brakepoints.
```
gabriel-glo, line 97:
```suggestion
   * @param array   $items        Array of breakpoints.
```

src/helpers/class-components.php
```diff
@@ -90,4 +90,37 @@ public static function render( string $component, array $attributes = [] ) {
...
+   * @param string  $selector     Selector for this brakepoint.
```
gabriel-glo, line 98:
```suggestion
   * @param string  $selector     Selector for this breakpoint.
```

PR #60, opened 2020-05-27
Fix phpstan issues

dingo-d:
I've had to exclude some checks as they are false positive. Most of the config calls are false-positive because they depend on the `Config_Data` interface, which dictates that those methods have to exist - but they will be implemented in the theme or a plugin, and there are no concrete implementations in the libs, so it says it's an error when it's not.

Walker has core inconsistencies (https://core.trac.wordpress.org/ticket/50261) that are throwing false-positives.

I've made some changes so it would be great if we test this before merging 😬 

Ideally, we'd have automated tests but I need to work on those...

### DIFF
--- a/src/class-main.php
+++ b/src/class-main.php
@@ -210,7 +210,7 @@ abstract class Main implements Service {
    *
    * @throws Exception\Invalid_Service If the service is not valid.
    *
-   * @return Service
+   * @return Registrable
    *
    * @since 0.1.0
    */
--- a/src/class-main.php
+++ b/src/class-main.php
@@ -144,7 +144,7 @@ abstract class Main implements Service {
     $output = [];
 
     foreach ( $this->get_service_classes() as $class => $dependencies ) {
-      if ( is_array( $dependencies ) ) {
+      if ( is_array( $dependencies ) ) { /** @phpstan-ignore-line */
         $output[ $class ] = $dependencies;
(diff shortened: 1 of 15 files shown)


iruzevic:
this looks good, please merge latest develop branch to your branch to resolve conflicts

szepeviktor:
> Call to an undefined method object::get_config().

Those are easy to fix!

If you merge this one I'll send a PR right away.

src/class-main.php
```diff
@@ -144,7 +144,7 @@ private function get_service_classes_prepared_array() : array {
     $output = [];
 
     foreach ( $this->get_service_classes() as $class => $dependencies ) {
-      if ( is_array( $dependencies ) ) { /** @phpstan-ignore-line */
+      if ( is_array( $dependencies ) ) { /** @phpstan-ignore-line phpcs:ignore */
```
szepeviktor, line 147: Just set the return type of `get_service_classes`

`array<class-string, string|string[]>`

dingo-d: I had them set to just `array<string>` which caused errors. Thanks for the tip 👍

dingo-d:
@szepeviktor You can make the PR if you want with further fixes 👍

szepeviktor:
🏎️ Starting **now**...

PR #61, opened 2020-05-27
updating helper

iruzevic:
* Components:render helper was not working in Plugin Boilerplate setup

### DIFF
--- a/src/helpers/class-components.php
+++ b/src/helpers/class-components.php
@@ -55,22 +55,25 @@ class Components {
    *
    * @param  string $component  Component's name or full path (ending with .php).
    * @param  array  $attributes Array of attributes that's implicitly passed to component.
+   * @param  string $parentPath If parent path is provides it will be appended to the file location, if not get_template_directory_uri() will be used as a default parent path.
    * @return string
    *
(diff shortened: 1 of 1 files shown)


src/helpers/class-components.php
```diff
@@ -55,22 +55,25 @@ public static function classnames( array $classes ) : string {
...
-  public static function render( string $component, array $attributes = [] ) {
+  public static function render( string $component, array $attributes = [], string $parentPath = '' ) {
+
+    if ( empty( $parentPath ) ) {
+      $parentPath = get_template_directory_uri();
+    }
 
     // Detect if user passed component name or path.
     if ( strpos( $component, '.php' ) !== false ) {
       $component_path = $component;
```
IvanGrginovInf, line 71: This should probably be like this, no? (As to not break if somebody is already using `render` by providing a relative path from theme / plugin root folder)

```suggestion
      $component_path = "{$parentPath}/$component";
```

iruzevic: agreed

iruzevic: fixed

PR #63, opened 2020-06-03
Taxonomies for multiple post types

IvanGrginovInf:
Added
---
* Allow the same taxonomy to be registered for multiple post types.


IvanGrginovInf:
> This is a breaking change
Oh damn, you're absolutely right, because I've changed the typehint of the abstract method. We should leave this for `4.0.0` then.

iruzevic:
Implemented

PR #66, opened 2020-06-04
Fix/phpstan max

szepeviktor:
@dingo-d Please help me mitigate undefined `Manifest_Data::get_config()` - the last obsticle

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "afb9223f4c57883939a664506583646a",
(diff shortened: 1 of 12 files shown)


dingo-d:
Adding

```php
  /**
   * Config getter
   *
   * @since 2.2.0 Added config getter.
   *
   * @return Config_Data
   */
  public function get_config();
```

To `interface-manifest-data.php` (with appropriate `use` statement `use Eightshift_Libs\Core\Config_Data;`), and removing `object` from the return value of the `get_config()` method in the `class-manifest.php` seems to fix the `Call to an undefined method Eightshift_Libs\Manifest\Manifest_Data::get_config` issue.

szepeviktor:
Done 🍏

szepeviktor:
Thank you for providing me a task to work on.

dingo-d:
Thank you for helping us out 🙂  We really appreciate it.

PR #67, opened 2020-06-08
Add console generate command

dingo-d:
I've added the REST generation command. The plan is to add:

- Custom post type creation command 
- Custom taxonomy creation command 
- Custom block creation command 

basically, anything that implements a template method design pattern.

What this does is adds a `bin/console` script that, when invoked, will show this

Screenshot 2020-06-08 at 21 10 34

The added custom command will take 2 arguments: `endpoint-name` - the slug of the endpoint and `verb` - the HTTP verb for the rest route.

Based on the name, a class name file and class name will be generated.

Screenshot 2020-06-08 at 21 12 32

And if you don't provide correct arguments, an error will be thrown

Screenshot 2020-06-08 at 21 13 02

And you can see what is doing what by typing `bin/console generate:route --help`

Screenshot 2020-06-08 at 21 13 42


src/commands/class-generate-rest-route.php
```diff
@@ -0,0 +1,203 @@
...
+   *
+   * @param string $class_name Name of the REST class to create.
+   * @param string $endpoint   Name of the endpoint of the REST class.
+   * @param string $verb       HTTP verb.
+   *
+   * @return string Class boilerplate contents.
+   */
+  private function get_class_boilerplate( string $class_name, string $endpoint, string $verb ) {
+    return <<<EOT
+<?php
```
gabriel-glo, line 108: This code is commented out because this is WIP?

src/commands/class-generate-rest-route.php
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Class Generate_Rest_Route
+ *
+ * Symfony command generator class used for REST route generation.
+ *
+ * @package Eightshift_Libs\Commands
+ */
+class Generate_Rest_Route extends Command {
+
+  protected static $defaultName = 'generate:route'; // phpcs:ignore
```
IvanGrginovInf, line 28: Which rule is this ignoring? (No docblock?)

src/commands/class-generate-rest-route.php
```diff
@@ -0,0 +1,203 @@
...
+    ];
+
+    $class_boilerplate = $this->get_class_boilerplate( $class_name, $endpoint, $verb_mapping[ $http_verb ] );
+
+    $rest_dir  = dirname( __FILE__, 2 ) . '/rest';
+    $directory = $rest_dir . "/class-{$endpoint}.php";
+
+    $fp = fopen( $directory, 'wb' ); // phpcs:ignore
+    fwrite( $fp, $class_boilerplate ); // phpcs:ignore
+    fclose( $fp ); // phpcs:ignore
```
IvanGrginovInf, line 90: That's so many blanket phpcs:ignores 😄 Could you please put in an ignore for a specific rule?

src/commands/class-generate-rest-route.php
```diff
@@ -0,0 +1,203 @@
...
+   * Class boilerplate
+   *
+   * @param string $class_name Name of the REST class to create.
+   * @param string $endpoint   Name of the endpoint of the REST class.
+   * @param string $verb       HTTP verb.
+   *
+   * @return string Class boilerplate contents.
+   */
+  private function get_class_boilerplate( string $class_name, string $endpoint, string $verb ) {
+    return <<<EOT
```
IvanGrginovInf, line 107: We should probably have the Route template code in an example file (which you'd then just copy) rather than like this.

src/commands/class-generate-rest-route.php
```diff
@@ -0,0 +1,203 @@
...
+   * Configures the current command
+   *
+   * @inheritDoc
+   */
+  protected function configure() {
+    $this
+      ->setDescription( 'Generate a custom REST class' )
+      ->setHelp( 'This command will create a template class for the custom WordPress REST API endpoint.' )
+      ->addArgument( 'endpoint-name', InputArgument::REQUIRED, 'Pass the slug of your endpoint. Example: my-endpoint' )
+      ->addArgument( 'verb', InputArgument::REQUIRED, 'Pass the desired HTTP verb. Can be one of: \'GET\', \'POST\', \'PATCH\', \'PUT\', or \'DELETE\'' );
```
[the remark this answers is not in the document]
dingo-d: Because it's semantically correct (HTTP verb). I mean we can change it, I just went with what I thought was more in line with what REST request does 🤷‍♂️

src/commands/class-generate-rest-route.php
dingo-d: That's a [heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc).

src/commands/class-generate-rest-route.php
dingo-d: This is ignoring the `$defaultName`. WPCS expects `$default_name`, but this has to be such because it overrides a property in the `Command` class from Symfony. I'll add a comment in the `phpcs:ignore` about what it silences 👍

src/commands/class-generate-rest-route.php
dingo-d: I'll add rules it silences. In this case it's: use wp filesystem instead :v Which is... no XD

src/commands/class-generate-rest-route.php
dingo-d: I thought about this but kinda thought that would leave a lot of unused files laying around 🤷‍♂️  This felt a bit easier (plus replacing names, class names etc. looks a bit easier than reading files, replacing then saving them).

src/commands/class-generate-rest-route.php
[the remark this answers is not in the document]
gabriel-glo: I agree that this is semantically correct term, but I'm mostly referring to the general usage of these verbs. Usually we are using the term `method` when talking about GET, POST, etc.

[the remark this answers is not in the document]
dingo-d: I see now in the interface that we have: `Alias for GET transport method.` so `method` can be used 👍

iruzevic:
@dingo-d  can you please pull the latest master branch and fix this conflict so we can merge this?

src/commands/class-generate-rest-route.php
IvanGrginovInf: As you wish, this was just reminding me of a certain legacy project we've worked on.

dingo-d:
Ok, I'll try to extract this to a template file, and see how maker bundle does the search-replace when creating things like controllers etc.

dingo-d:
I'd merge this, but I need an approval from @gabriel-glo (who is on vacation)

PR #73, opened 2020-07-09
adding setup script for deployment

iruzevic:
This works!!!
You just run `sh bin/setup.sh`.

Screenshot 2020-07-09 at 16 19 12


iruzevic:
> While this is great, I'm not happy about the jq dependency. It's not usually added to servers so it's an extra thing that we'll need to ask the devOps to add either to Semaphore or on servers.
> 
> Could we use PHP to run these?

I will try

PR #77, opened 2020-07-19
Feature/di container

iruzevic:
### Changed
- Switching from Classmap to PSR-4 naming.
- example: `class-media.php` is now `Media.php`.
- example: `menu.php` is now `AbstractMenu.php`.
- example: `interface-manifest-data.php` is now `ManifestDataInterface.php`.
- example: `trait-error-logger.php` is now `ErrorLoggerTrait.php`.
- example: `class-component-exception.php` is now `FinalComponentException.php`.
- replaced `DI\create` with `new AutowireDefinitionHelper();`.
- replaced `DI\get` with `new Reference( $dependency );`.
- changed namespace `Eightshift_Libs` to `EightshiftLibs`.

### Removed
- removed `instantiate_service()` method.


bin/console
```diff
@@ -2,7 +2,7 @@
 <?php
 
 use Symfony\Component\Console\Application;
-use Eightshift_Libs\Commands\Generate_Rest_Route;
+use EightshiftLibs\Commands\Generate_Rest_Route;
```
dingo-d, line 5: Shouldn't this be `GenerateRestRoute`, not `Generate_Rest_Route`?

composer.json
```diff
@@ -27,8 +27,7 @@
   "require": {
     "php": ">=7.2",
     "ext-json": "*",
-    "php-di/php-di": "^6.0",
-    "symfony/console": "^5.1"
+    "php-di/php-di": "dev-master"
```
dingo-d, line 30: We shouldn't depend on `master` branch. We should see when the never version will come out and update to that version instead.

composer.json
```diff
@@ -27,8 +27,7 @@
...
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.6",
```
dingo-d, line 33: 0.7 version has come out with some improvements so we can update this as well

src/AbstractMain.php
```diff
@@ -2,34 +2,30 @@
...
 /**
  * The main start class.
  * This is used to define instantiate all classes used in the lib.
  *
- * @since 2.0.0 Adding project prefix constant to use in lib.
- * @since 0.7.0 Dependency Injection Refactoring.
- * @since 0.1.0
  */
-abstract class Main implements Service {
+abstract class AbstractMain implements ServiceInterface {
```
dingo-d, line 22: I know about the idea of prefixing and suffixing class names based on what they are, but `AbstractMain` just sounds so abstract (no pun intended 😂 ).

It doesn't convey at all what this class (in this case a template) does and why you should extend it. I'm trying to figure out the good name, but it's hard (naming is hard lol).
Since this is the main class where you register all the services, something like `ServiceRegistrator` or something like that?

Suggestions, discussion etc. are welcomed 🙂

src/AbstractMain.php
```diff
@@ -165,21 +149,22 @@ private function get_service_classes_prepared_array() : array {
...
-
     $definitions = [];
+
     foreach ( $services as  $service_key => $service_values ) {
       if ( gettype( $service_values ) !== 'array' ) {
         continue;
       }
 
-      $definitions[ $service_key ] = \DI\create()->constructor( ...$this->get_di_dependencies( $service_values ) );
+      $autowire = new AutowireDefinitionHelper();
```
dingo-d, line 161: Maybe we should add this as a dependency?

```php
use DI\Definition\Helper\DefinitionHelper;

public function __construct(DefinitionHelper $definitionHelper) {
  $this->definitionHelper = $definitionHelper;
}
```

And then pass this as a dependency? The same is with `ContainerBuilder`.

It would be ideal that we actually have a `DiContainer` where we can extract all the dependency injection logic, and handle dependencies there.

Since we are making a breaking change, let's do some improvements as well 🙂

src/blocks/Blocks.php
```diff
@@ -331,15 +301,13 @@ public function render_wrapper_view( string $src, array $attributes, $inner_bloc
...
-   *
-   * @since 2.0.0
+   * @throws FinalInvalidBlock Throws error if render block view is missing.
    */
   public function render_block_view( string $src, array $attributes, $inner_block_content = null ) : void {
     $path = $this->get_blocks_path() . $src;
 
     if ( ! file_exists( $path ) ) {
-      throw Invalid_Block::missing_render_view_exception( $path );
+      throw FinalInvalidBlock::missing_render_view_exception( $path );
```
dingo-d, line 310: Just a question: you named `FinalInvalidBlock` just because a class is `final`? That's like really unnecessary. Any IDE will warn a user if they can or cannot extend a class if its final or not. And in any case they'll get an error, and should read the documentation.

So IMO I think this naming things based on what they are is a bit too extreme...

composer.json
iruzevic: Yes this will be updated when the new version is out

src/AbstractMain.php
iruzevic: Sounds okej, We will discuss on Wp Standup.

src/customPostType/FinalLabelGenerator.php
```diff
@@ -4,64 +4,52 @@
...
-use Eightshift_Libs\Exception\Invalid_Nouns;
+use EightshiftLibs\Exception\FinalInvalidNouns;
 
 /**
  * Class that generates lables for custom post type.
- *
- * @since 0.1.0
  */
-final class Label_Generator {
+final class FinalLabelGenerator {
```
dingo-d, line 19: Same comment as with the exceptions, `Final` prefix really isn't needed

src/AbstractMain.php
iruzevic: First thing sure sounds good.

The other one no, I have changed ->get and ->create an ->autowire() with the actual classes. Because in PHP-DI these are only helper classes that are not inside a namespace so they create an issue with prefixing all vendor packages.

src/blocks/Blocks.php
iruzevic: I have followed the psr-4 naming recommendations. We can change this

dingo-d: Yeah, PSR-4 is for autoloading only, not naming conventions. Also, the class file would have to be `FinalInvalidBlock.php` in this case.

I'm kinda ok with `Abstract` and `Interface` or `Trait` suffix/prefix (although I do think sometimes this is an overkill), but putting `Final` just makes no sense to me 😬

iruzevic: Let's discuss this on WP Standup

iruzevic:
> Left some comments.
> 
> As with the boilerplate, we should see if we want to implement PSR-12. And we need to agree on naming, because some things are badly named (especially the final classes).

Sure, good idea. WP Standup :D

src/columns/AbstractBasePostTypeColumns.php
```diff
@@ -2,29 +2,26 @@
...
+use EightshiftLibs\Core\ServiceInterface;
 
 /**
- * Abstract class Base_Post_Type_Columns.
+ * Abstract class AbstractBasePostTypeColumns.
  *
  * This abstract class can be extended to add (custom) post columns in the post screen.
  */
-abstract class Base_Post_Type_Columns implements Service {
+abstract class AbstractBasePostTypeColumns implements ServiceInterface {
```
IvanGrginovInf, line 19: Similar to what Denis pointed out but specifically having both `Abstract` and `Base` in the name is redundant. We should stick to single one (and I personally feel we **should** prefix abstract Libs classes with one or the other).

src/console/ConsoleHelpers.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @package EightshiftLibs\Console
+ */
+
+declare( strict_types=1 );
+
+namespace EightshiftLibs\Console;
+
+/**
+ * Helpers for components
```
IvanGrginovInf, line 13:
```suggestion
 * Helpers for console
```

src/console/ConsoleHelpers.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Helpers for components
```
IvanGrginovInf, line 3:
```suggestion
 * Helpers for console
```

src/console/console.php
```diff
@@ -0,0 +1,23 @@
...
+use EightshiftLibs\Rest\Routes\GenerateRestRoute;
+use Symfony\Component\Console\Application;
+
+function consoleRun( string $root ) {
+
+  if (!in_array(PHP_SAPI, ['cli', 'phpdbg', 'embed'], true)) {
+    echo 'Warning: The console should be invoked via the CLI version of PHP, not the '.PHP_SAPI.' SAPI'.PHP_EOL;
+  }
+
+  set_time_limit(0);
```
IvanGrginovInf, line 12: You shouldn't really do that, a high value sure but not setting it to unlimited. Also this is only generating a rest route (a file), this shouldn't take more than couple of seconds at max, did you have issues with it?

dingo-d: This is out of the box by Symfony iirc. And since it's CLI it shouldn't be a problem I think.

src/blocks/AbstractBlocks.php
```diff
@@ -9,8 +9,7 @@
 declare( strict_types=1 );
 
 namespace EightshiftLibs\Blocks;
-
-use EightshiftLibs\Config\ConfigInterface;
+;
```
dingo-d, line 12: Typo

src/blocks/AbstractBlocks.php
```diff
@@ -426,13 +416,13 @@ protected function get_attributes( array $block_details ) : array {
       $block_details['attributes']
     );
 
-    $filter_name = $this->config->get_config( static::BLOCK_ATTRIBUTES_FILTER_NAME );
+    // $filter_name = $this->config->get_config( static::BLOCK_ATTRIBUTES_FILTER_NAME );
```
dingo-d, line 419: Comment left

src/cli/CliHelpers.php
```diff
@@ -301,4 +301,17 @@ public function get_vendor_prefix( array $args = [] ) : string {
 
     return $vendor_prefix;
   }
+
+  public function is_dependant( string $string ) {
+    var_dump($this->get_output_dir('main/Main.php'));
```
dingo-d, line 306: Still a WIP?

src/customTaxonomy/Taxonomy.php
```diff
@@ -22,10 +22,10 @@ class Taxonomy extends AbstractTaxonomy {
    * @return void
    */
   public function register() {
-    add_action(
+    \add_action(
       'init',
       function() {
-        register_taxonomy(
+        \register_taxonomy(
```
dingo-d: While I absolutely love anonymous functions, it makes code basically untestable and unhookable. We should move this to a separate public method.

src/rest/routes/AbstractRoute.php
```diff
@@ -23,10 +23,10 @@ abstract class AbstractRoute implements RouteInterface, ServiceInterface {
    * @return void
    */
   public function register() : void {
-    add_action(
+    \add_action(
       'rest_api_init',
       function() {
-        register_rest_route(
+        \register_rest_route(
```
dingo-d, line 29: Same as with the previous comment.

src/rest/routes/RouteCli.php
```diff
@@ -103,6 +103,13 @@ public function __invoke( array $args, array $assoc_args ) {
     // Remove unecesery stuff from props.
     $endpoint = str_replace( '_', '-', str_replace( ' ', '-', strtolower( $endpoint_slug ) ) );
 
+    // var_dump($class);
+
+    if (strpos( $class , 'ConfigInterface $config') !== false) {
+      var_dump('DAAAA');
```
dingo-d, line 109: For some reason, this reminded me of 'This is Spartaaaaa' 😂 

Gosh, I'm playing too much of Assassin's Creed Odyssey 😆

src/blocks/AbstractBlocks.php
iruzevic: 👍

src/blocks/AbstractBlocks.php
iruzevic: still WIP

src/cli/CliHelpers.php
iruzevic: yes

src/cli/Cli.php
```diff
@@ -128,7 +182,9 @@ public function run( string $command_parent_name ) {
     $this->command_parent_name = $command_parent_name;
 
     $this->run_command( new ConfigCli() );
+    $this->run_command( new TaxonomyCli() );
     $this->run_command( new I18nCli() );
+    $this->run_command( new LoginCli() );
     $this->run_command( new MainCli() );
     $this->run_command( new ManifestCli() );
     $this->run_command( new MediaCli() );
```
dingo-d, line 191: On line 204 you have 

```php
$class->__invoke(
  [],
  $args
);
```

if the class has `__invoke()` magic method, couldn't that be written something like `$register = new $class; $register([], $args)` ?

Because it's assumed that `$class` is invokable (has `__invoke()` magic method), so it can be called directly.

https://www.php.net/manual/en/language.oop5.magic.php#object.invoke

I'm not sure I understand what this does exactly.

src/cli/CliHelpers.php
```diff
@@ -21,6 +21,11 @@ trait CliHelpers {
    * @return string
    */
   public function get_file_name( string $file_name ) : string {
+
+    if ( strpos( $file_name, ' ' ) !== false) {
```
dingo-d, line 25:
```suggestion
    if ( strpos( $file_name, ' ' ) !== false ) {
```

dingo-d:
Just a comment, not to forget: we need to have a big: BC break notice in the changelog, since we've added type hints on methods, and just updating the libs would break it.

IvanGrginovInf (approved):

IvanGrginovInf (approved):

src/cli/CliHelpers.php
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  /**
+   * Generate example template file/class name.
+   *
+   * @param string $string File name
+   *
+   * @return string
+   */
+  public function get_example_file_name( $string ) : string {
```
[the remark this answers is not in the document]
iruzevic: We can fine tune that later

src/cli/Cli.php
iruzevic: This has changes drastically so this comment doesn't apply anymore.

PR #81, opened 2020-09-11
Autowiring

IvanGrginovInf:
Working implementation of autowiring for both interfaces and classes (+ Codeception unit tests setup). 🎉 

What it does
---
* It automatically builds definition lists for dependencies for all services and other classes which have dependencies. (This means no more manually adding classes to `class-main.php`)
* It injects class-based dependencies
* It injects interface-based dependencies based on variable name. How? It will look through all psr-4 compliant classes and try to find exactly 1 by matching the dependency variable name to classname (and seeing if that class implements the desired interface). If it finds 0 or 2+ classes, it will just skip it and you'll have to manually define dependencies for that class.
* Services are automatically put on the definition list and then registered as usual

This is a breaking change for several reasons:
---
* It only works with psr-4
* It adds a dependency to `class-main.php` (`class-autowire.php` which itself has a dependency of `Composer's ClassLoader->getPsr4Prefixes()`) (meaning we need to change `functions.php` in both theme & plugin).

Here's an example of `functions.php` needed to get this working:

```php
/**
 * Include the autoloader so we can dynamically include the rest of the classes.
 *
 * @since 1.0.0
 */
$loader = require __DIR__ . '/vendor/autoload.php';

/**
 * Begins execution of the theme.
 *
 * Since everything within the theme is registered via hooks,
 * then kicking off the theme from this point in the file does
 * not affect the page life cycle.
 *
 * @since 1.0.0
 */
( new Core\Main( new Core\Autowiring( $loader->getPrefixesPsr4() ) ) )->register();

```

To Do
---
- [ ] - Fix code to match new coding standards once those are in (currently it's a mess)
- [ ] - Setup unit test. Currently I've just tested it on a project with some dummy dependencies. I'd like to build an all-encompassing unit test with all possible variations.
- [ ] - (Maybe) add some optimisation (since this runs on each request). Currently the whole autowiring process lasts **between 0.012s and 0.005s with ~10 classes** which is fairly fast. I haven't benchmarked it with a larger number of classes.

### DIFF
--- a/src/class-autowiring.php
+++ b/src/class-autowiring.php
@@ -1,0 +1,264 @@
+<?php
+/**
+ * The file that defines the autowiring process
+ *
+ * @package Eightshift_Libs\Core
+ */
+
+declare( strict_types=1 );
+
+namespace Eightshift_Libs\Core;
+
+/**
+ * The file that defines the autowiring process
+ */
+class Autowiring {
+
+  /**
+   * Constructs object and inserts prefixes from composer.
+   *
+   * @param array $psr4Prefixes Composer's ClassLoader psr4Prefixes. $ClassLoader->getPsr4Prefixes().
+   */
+  public function __construct( array $psr4Prefixes ) {
+    $this->psr4Prefixes = $psr4Prefixes;
+  }
+
+  /**
+   * Autowiring.
+   *
+   * @return array<array> Array of fully qualified class names.
+   */
+  public function buildServiceClasses() : array {
+    $start = microtime( true );
+    $projectClasses = $this->getClassesInNamespace( 'AutowiringTest', $this->psr4Prefixes );
+
+    $dependencyTree = [];
+    $filenameIndex  = [];
+
+    // Prepare the filename index.
+    $filenameIndex       = $this->buildFilenameIndex( $projectClasses );
+    $classInterfaceIndex = $this->buildClassInterfaceIndex( $projectClasses );
+
+    foreach ( $projectClasses as $projectClass ) {
+      $reflClass = new \ReflectionClass( $projectClass );
+
+      // Skip abstract classes, interfaces & traits.
+      if ( $reflClass->isAbstract() || $reflClass->isInterface() || $reflClass->isTrait() ) {
+        continue;
+      }
+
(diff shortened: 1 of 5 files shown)


iruzevic:
Please make a PR to release/3.0.0 branch and fixe namespacing and also put this file inside src/Main folder and rename it to PSR-4 standard

src/class-autowiring.php
```diff
@@ -0,0 +1,284 @@
...
+   * @param array $psr4Prefixes Composer's ClassLoader psr4Prefixes. $ClassLoader->getPsr4Prefixes().
+   */
+  public function __construct( array $psr4Prefixes ) {
+    $this->psr4Prefixes = $psr4Prefixes;
+  }
+
+  /**
+   * Autowiring.
+   *
+   * @return array<array> Array of fully qualified class names.
```
gabriel-glo, line 36:
```suggestion
   * @return array Array of fully qualified class names.
```

iruzevic (changes requested):
Please make a PR to release/3.0.0 branch and fixe namespacing and also put this file inside src/Main folder and rename it to PSR-4 standard

src/class-autowiring.php
```diff
@@ -0,0 +1,284 @@
...
+      // Skip irrelevant classes.
+      if (
+        ! $reflClass->implementsInterface( 'Eightshift_Libs\Core\Service' )
+        && ( empty( $reflClass->getConstructor() ) || empty( $reflClass->getConstructor()->getParameters() ) )
+      ) {
+        continue;
+      }
+
+      // Build the dependency tree.
+      $dependencyTree = array_merge($this->buildDependencyTree($projectClass, $filenameIndex, $classInterfaceIndex), $dependencyTree);
```
gabriel-glo, line 65: Brackets spacing in multiple instances in this file.
Run PHP linter.

gabriel-glo (changes requested):
Just some smaller issues.

src/class-autowiring.php
IvanGrginovInf: No point to run phpcs yet since we're completely revamping standards. Also this is how psr-4 works out of the box.

PR #82, opened 2020-09-11
Autowiring

IvanGrginovInf:
Working implementation of autowiring for both interfaces and classes (+ Codeception unit tests setup). 🎉 

What it does
---
* It automatically builds definition lists for dependencies for all services and other classes which have dependencies. (This means no more manually adding classes to `class-main.php`)
* It injects class-based dependencies
* It injects interface-based dependencies based on variable name. How? It will look through all psr-4 compliant classes and try to find exactly 1 by matching the dependency variable name to classname (and seeing if that class implements the desired interface). If it finds 0 or 2+ classes, it will just skip it and you'll have to manually define dependencies for that class.
* Services are automatically put on the definition list and then registered as usual

This is a breaking change for several reasons:
---
* It only works with psr-4
* It adds a dependency to `class-main.php` (`class-autowire.php` which itself has a dependency of `Composer's ClassLoader->getPsr4Prefixes()`) (meaning we need to change `functions.php` in both theme & plugin).

Here's an example of `functions.php` needed to get this working:

```php
/**
 * Include the autoloader so we can dynamically include the rest of the classes.
 *
 * @since 1.0.0
 */
$loader = require __DIR__ . '/vendor/autoload.php';

/**
 * Begins execution of the theme.
 *
 * Since everything within the theme is registered via hooks,
 * then kicking off the theme from this point in the file does
 * not affect the page life cycle.
 *
 * @since 1.0.0
 */
( new Main\Main( new Main\Autowiring( $loader->getPrefixesPsr4() ) ) )->register();

```

To Do
---
- [ ] Fix code to match new coding standards once those are finished (currently it's a mess)
- [ ] Setup unit test. Currently I've just tested it on a project with some dummy dependencies. I'd like to build an all-encompassing unit test with all possible variations.
- [ ] (Maybe) add some optimisation (since this runs on each request). Currently the whole autowiring process lasts **between 0.012s and 0.005s with ~10 classes** which is fairly fast. I haven't benchmarked it with a larger number of classes.

### DIFF
--- a/src/Main/Autowiring.php
+++ b/src/Main/Autowiring.php
@@ -1,0 +1,264 @@
+<?php
+/**
+ * The file that defines the autowiring process
+ *
+ * @package Eightshift_Libs\Main
+ */
+
+declare( strict_types=1 );
+
+namespace Eightshift_Libs\Main;
+
+/**
+ * The file that defines the autowiring process
+ */
+class Autowiring {
+
+  /**
+   * Constructs object and inserts prefixes from composer.
+   *
+   * @param array $psr4Prefixes Composer's ClassLoader psr4Prefixes. $ClassLoader->getPsr4Prefixes().
+   */
+  public function __construct( array $psr4Prefixes ) {
+    $this->psr4Prefixes = $psr4Prefixes;
+  }
+
+  /**
+   * Autowiring.
+   *
+   * @return array<array> Array of fully qualified class names.
+   */
+  public function buildServiceClasses() : array {
+    $start = microtime( true );
+    $projectClasses = $this->getClassesInNamespace( 'AutowiringTest', $this->psr4Prefixes );
+
+    $dependencyTree = [];
+    $filenameIndex  = [];
+
+    // Prepare the filename index.
+    $filenameIndex       = $this->buildFilenameIndex( $projectClasses );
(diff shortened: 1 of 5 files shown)


src/Main/Autowiring.php
```diff
@@ -0,0 +1,284 @@
...
+      $reflClass = new \ReflectionClass( $projectClass );
+
+      // Skip abstract classes, interfaces & traits.
+      if ( $reflClass->isAbstract() || $reflClass->isInterface() || $reflClass->isTrait() ) {
+        continue;
+      }
+
+      // Skip irrelevant classes.
+      if (
+        ! $reflClass->implementsInterface( 'Eightshift_Libs\Core\Service' )
```
iruzevic, line 58: what does this do?

we need to check how this will work with vendor prefixing

dingo-d:
Can you fix the PHPCS and PHPStan issues @IvanGrginovInf ?

gabriel-glo (approved):

iruzevic:
> Can you fix the PHPCS and PHPStan issues @IvanGrginovInf ?

We will do this once we merge new standards

dingo-d:
PHPStan issues can be fixed regardless of the standards 😉

PR #83, opened 2020-09-14
Feature/autowiring release 3

iruzevic:
I have:

- [ ] fixed linting issues
- [ ] Implemented fix to exclude view files and non php files
- [ ] fixed some issues with missing names and etc.

This is tested on actual project with BP 5.0.0 and it works

### DIFF
--- a/src/Main/AbstractMain.php
+++ b/src/Main/AbstractMain.php
@@ -100,7 +100,7 @@ abstract class AbstractMain implements ServiceInterface {
    * @return array<array>
    */
   private function get_service_classes_with_autowire() {
-    return array_merge( $this->autowiring->buildServiceClasses(), $this->get_service_classes() );
+    return array_merge( $this->autowiring->build_service_classes(), $this->get_service_classes() );
   }
 
   /**
--- a/src/Main/AbstractMain.php
+++ b/src/Main/AbstractMain.php
@@ -100,7 +100,7 @@ abstract class AbstractMain implements ServiceInterface {
    * @return array<array>
    */
   private function get_service_classes_with_autowire() {
-    return array_merge( $this->autowiring->build_service_classes(), $this->get_service_classes() );
+    return array_merge( $this->autowiring->buildServiceClasses(), $this->get_service_classes() );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/Main/AbstractMain.php
```diff
@@ -100,7 +100,7 @@ public function build_di_container() {
    * @return array<array>
    */
   private function get_service_classes_with_autowire() {
-    return array_merge( $this->autowiring->buildServiceClasses(), $this->get_service_classes() );
+    return array_merge( $this->autowiring->build_service_classes(), $this->get_service_classes() );
```
dingo-d, line 103: PSR-4 it should be `buildServiceClasses()` 🤔

dingo-d (commented):
Why did you replace method names to snake_case instead of camelCase?

src/Main/AbstractMain.php
iruzevic: fixed

iruzevic:
> Why did you replace method names to snake_case instead of camelCase?

Didn't knew about the new psr-12 standard. It is updated

src/Main/Autowiring.php
```diff
@@ -1,335 +1,337 @@
...
+		if (! is_dir($pathToNamespace)) {
+			return [];
+		}
+
+		$it = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($pathToNamespace));
+		foreach ($it as $file) {
+			if ($file->isDir()) {
+				continue;
+			}
+			if (preg_match('/\.(php)$/', $file->getFileName()) && preg_match('/^[A-Z]{1,}.*$/m', $file->getFileName())) {
```
IvanGrginovInf, line 192: I think you should be able to do it as a single regex experssion like so (all filename starting with capital letter, ending with `.php`)

```suggestion
			if (preg_match('/[A-Z]{1}.*.php/', $file->getFileName())) {
```

IvanGrginovInf (changes requested):
Left a single comment regarding simplifying regex, otherwise looks good.

src/Main/Autowiring.php
iruzevic: Tnx I have implemented the fix

IvanGrginovInf (approved):

IvanGrginovInf:
@iruzevic Approved but please fix the conflict before mering.

PR #84, opened 2020-09-15
Fix lint issues

dingo-d:
This is the first pass of the new linter. I fixed all of the linting issues, plus added some exceptions in the `phpcs.xml.dist`.

After this, I'll go through the PHPStan and fix the issues. I've noticed some methods that should maybe be private, some docblocs are not correct, or the type hints are missing, etc.

There are a lot of changes, but it's the only way to test the standards. Let me know if you notice anything out of place, something that was broken during the `phpcbf`, or that I missed when manually fixing the rest of the issues.

### DIFF
--- a/src/Blocks/AbstractBlocks.php
+++ b/src/Blocks/AbstractBlocks.php
@@ -151,12 +151,9 @@ abstract class AbstractBlocks implements ServiceInterface, RenderableBlockInterf
 			throw InvalidBlock::missingBlocksException();
 		}
 
-		\array_map(
-			function ($block) {
-				$this->registerBlock($block);
-			},
-			$blocks
-		);
+		foreach ($blocks as $block) {
+			$this->registerBlock($block);
+		}
 	}
 
 	/**
--- a/src/Blocks/AbstractBlocks.php
+++ b/src/Blocks/AbstractBlocks.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Class Blocks is the base class for Gutenberg blocks registration.
  * It provides the ability to register custom blocks using manifest.json.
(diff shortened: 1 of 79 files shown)


src/Blocks/BlockCli.php
```diff
@@ -49,7 +53,14 @@ public function getDoc() : array {
 		];
 	}
 
-	public function __invoke( array $args, array $assocArgs ) { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing, Generic.CodeAnalysis.UnusedFunctionParameter.FoundInExtendedClassBeforeLastUsed
+	/**
+	 * Creates a block
+	 *
+	 * @param array $args      Array of arguments form terminal.
+	 * @param array $assocArgs Array of associative arguments form terminal.
+	 */
```
iruzevic, line 61: This method can't have doc block. Because this will automatically be shown in the WP CLI terminal as a command description and we don't want that because we have getDoc() method for this.

Please remove this

iruzevic: and on every __invoke method

iruzevic (changes requested):
one change but everything is ook

src/Blocks/AbstractBlocks.php
```diff
@@ -136,21 +143,20 @@ function( $block ) {
...
-				},
-				$blocks
-			);
-		}
+		\array_map(
+			function ($block) {
+				$this->registerBlock($block);
+			},
+			$blocks
+		);
```
IvanGrginovInf, line 159: Should be `foreach` because we're not returning anything.

IvanGrginovInf (commented):
Left a comment regarding `array_map` otherwise looks awesome! I for one welcome our PSR-12 overlords 😄

src/Blocks/BlockCli.php
dingo-d: Oh, that's unfortunate :S I'll revert this

iruzevic (approved):

PR #85, opened 2020-09-15
Phpstan fixes

dingo-d:
I'm going folder by folder and fixing PHPStan issues and all PHPCS issues I find.

Some things are not clear to me, so I won't change them.


src/Columns/PostType/AbstractPostTypeColumns.php
```diff
@@ -28,7 +28,7 @@ abstract class AbstractPostTypeColumns implements ServiceInterface
 	public function register(): void
 	{
 		array_map(
-			function ($post_type ) {
+			function ($post_type) {
 				add_filter("manage_{$post_type}_posts_columns", [ $this, 'addColumnName' ]);
 				add_action("manage_{$post_type}_posts_custom_column", [ $this, 'renderColumnContent' ], 10, 2);
 			},
```
IvanGrginovInf, line 34: This isn't what `array_map` is meant to be used for (map is used when you wish to map **each member** of old array to a new array. Here we're not returning anything). Let's stick to the good old `foreach` for these kinds of things (instead of shoehorning these functions).

src/Columns/Taxonomy/AbstractTaxonomyColumns.php
```diff
@@ -28,7 +28,7 @@ abstract class AbstractTaxonomyColumns implements ServiceInterface
 	public function register(): void
 	{
 		array_map(
-			function ($taxonomy ) {
+			function ($taxonomy) {
 				add_filter("manage_edit-{$taxonomy}_columns", [ $this, 'addColumnName' ]);
 				add_filter("manage_{$taxonomy}_custom_column", [ $this, 'renderColumnContent' ], 10, 3);
 			},
```
IvanGrginovInf, line 34: Same as comment above, `array_map` should be replaced with `foreach`.

IvanGrginovInf (commented):
Left 2 comments regarding `array_map`, everything else looks good.

src/Columns/PostType/AbstractPostTypeColumns.php
iruzevic: Ignore this columns files they are not refactored yet

src/Columns/Taxonomy/AbstractTaxonomyColumns.php
iruzevic: the same you can ignore it

src/Menu/BemMenuWalker.php
```diff
@@ -66,13 +65,18 @@ public function __construct(string $cssClassPrefix)
...
-	public function display_element($element, &$children_elements, $max_depth, $depth = 0, $args, &$output) // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps, PEAR.Functions.ValidDefaultValue.NotAtEnd
-	{
-
+	public function display_element( // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps, PEAR.Functions.ValidDefaultValue.NotAtEnd
+		$element,
+		&$children_elements,
+		$max_depth, // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps, PEAR.Functions.ValidDefaultValue.NotAtEnd
+		$depth = 0,
+		$args, // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps, PEAR.Functions.ValidDefaultValue.NotAtEnd
+		&$output
```
IvanGrginovInf, line 74: Nothing wrong here, just curious, this is a PSR-12 thing or PHPStan thing or something? Like the line is too long so we need to break it up?

src/Menu/BemMenuWalker.php
```diff
@@ -66,13 +65,18 @@ public function __construct(string $cssClassPrefix)
...
+	) {
 		$id_field = $this->db_fields['id'];
 
-		if (is_object($args[0])) {
-			$args[0]->has_children = ! empty($children_elements[ $element->$id_field ]);
+		if (is_object($args[0]) && property_exists($args[0], 'has_children')) {
```
IvanGrginovInf, line 78: Btw you can just do this without raising any warning (it's probably a lot faster):
```suggestion
		if (isset($args[0]-> has_children)) {
```
Source: https://www.php.net/manual/en/function.isset.php#86313

The only difference with isset is that it return FALSE if `has_children == null` but I don't think this is relevant in this case. (source: https://www.php.net/manual/en/function.property-exists.php)

IvanGrginovInf (commented):
Left a question and a suggestion, otherwise looks good!

src/Menu/BemMenuWalker.php
dingo-d: Too many arguments to fit in one line so I split it up. We can revert this and just ignore any errors.

src/Menu/BemMenuWalker.php
dingo-d: Will check what PHPStan says 😄

iruzevic (approved):

PR #86, opened 2020-09-30
Codeception test setup

IvanGrginovInf:
Done
---
* Setup WP Browser
* Setup an example unit test
* Added unit tests to workflow


iruzevic (approved):

.env.testing
```diff
@@ -0,0 +1,18 @@
+TEST_SITE_DB_DSN=mysql:host=localhost;dbname=test
```
dingo-d, line 1: I'm 99% sure this won't work with the integration test setup on GH actions, but we'll cross that bridge when we get there 😄

IvanGrginovInf: Yeah, it won't unfortunately. Just a placeholder as setup as default by `wpbrowser`.

tests/unit/Helpers/ComponentsTest.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    /**
+     * @var \UnitTester
+     */
+    protected $tester;
+
+    public function testEnsureString()
+    {
+        $this->assertIsString(Components::ensureString('asdad'));
+        $this->assertIsString(Components::ensureString(['aaa', 'bbb']));
```
dingo-d, line 15: Try to test the 'unhappy path' as well. Like, what would happen if you pass some totally bonkers thing to `Components::ensureString`? Will it throw an error, or will try to turn that into a string, but return false? Will it completely break the app?

dingo-d (changes requested):
One small comment for the test. Good work!

tests/unit/Helpers/ComponentsTest.php
IvanGrginovInf: Found a bug in my function because of this comment, nice! 😄

PR #90, opened 2020-10-25
Run CI on Composer v2

szepeviktor:


iruzevic (approved):

.github/workflows/ci.yml
```diff
@@ -41,10 +41,10 @@ jobs:
                   ${{ runner.os }}-composer-
 
             - name: Install composer packages
-              run: composer install --no-progress
+              run: composer update --no-progress
```
dingo-d, line 44: Won't this update the packages?

.github/workflows/ci.yml
```diff
@@ -41,10 +41,10 @@ jobs:
...
 
             - name: Check coding standards using PHPCS
-              run: composer standards:check -- --runtime-set testVersion ${{ matrix.php }}-
+              run: composer run-script standards:check -- --runtime-set testVersion ${{ matrix.php }}-
```
dingo-d, line 47: The script runs without the `run-script` addition. Is there any benefit to adding it?

dingo-d (commented):
I've left some comments.

.github/workflows/ci.yml
szepeviktor: @dingo-d As we do not have a lock file, there is no way to have fixed versions.
BTW Having a lock file a really bad idea: e.g. testing on different PHP version etc.
https://github.com/ergebnis/php-library-template/blob/main/.github/workflows/integrate.yaml#L28-L29
Testing could be done with lowest and normal (highest) dependencies.

.github/workflows/ci.yml
szepeviktor: @dingo-d As we do not have a lock file, there is no way to have fixed versions.
BTW Having a lock file a really bad idea: e.g. testing on different PHP version etc.
https://github.com/ergebnis/php-library-template/blob/main/.github/workflows/integrate.yaml#L28-L29
Testing could be done with lowest and normal (highest) dependencies.

.github/workflows/ci.yml
szepeviktor: Writing explicit code aims review.
I am not the one to convince you :)

szepeviktor: e.g. do you understand `grep -rFnwh`? :-D

dingo-d: Yeah, but running `composer` will show extra scripts 😄  I mean, a bit more verbosity doesn't hurt 😄

dingo-d (approved):

PR #91, opened 2020-10-25
Remove unused *private* instantiate_service method

szepeviktor:
Please see https://github.com/infinum/eightshift-libs/pull/90/checks?check_run_id=1305354674 and https://github.com/infinum/eightshift-libs/search?q=instantiate_service

### DIFF
--- a/src/class-main.php
+++ b/src/class-main.php
@@ -205,30 +205,6 @@ abstract class Main implements Service {
   }
 
   /**
-   * Instantiate a single service.
-   *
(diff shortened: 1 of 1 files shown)


iruzevic (approved):

dingo-d (approved):
I think we removed that in https://github.com/infinum/eightshift-libs/blob/release/3.0.0/src/Main/AbstractMain.php

Thanks for the fix 👍

szepeviktor:
You're welcome.

PR #103, opened 2020-11-30
Allowed more attributes on buttons

gabriel-glo:
Allowed class, id and type attributes to buttons rendered by `Components::render` while extending `AbstractEscapedView` class

This broke the hamburger and drawer components as hamburger button didn't have class attribute.

### DIFF
--- a/src/View/AbstractEscapedView.php
+++ b/src/View/AbstractEscapedView.php
@@ -149,7 +149,10 @@ abstract class AbstractEscapedView implements ServiceInterface
 			'allowfullscreen' => true,
 		],
 		'button' => [
-			'onClick' => true,
+      'class' => true,
+      'id' => true,
+			'type' => true,
+      'onClick' => true,
 		],
 	];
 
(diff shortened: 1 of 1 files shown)


gabriel-glo:
@dingo-d All these test failed because they tried to access sniff called Infinum and I think you renamed that.
Do these tests need to be updated with the name change to Eightshift?

src/View/AbstractEscapedView.php
```diff
@@ -149,7 +149,10 @@ abstract class AbstractEscapedView implements ServiceInterface
 			'allowfullscreen' => true,
 		],
 		'button' => [
-			'onClick' => true,
+      'class' => true,
+      'id' => true,
+			'type' => true,
+      'onClick' => true,
```
IvanGrginovInf, line 155: You're mixing tabs & spaces with the rest of the tags here.

IvanGrginovInf (changes requested):

src/View/AbstractEscapedView.php
gabriel-glo: Didn't even notice. I need to change my editor settings.

iruzevic (approved):

volfkarlo (approved):

IvanGrginovInf (approved):

gabriel-glo:
Do you think we need to add even more attributes for the button element?

PR #104, opened 2020-12-01
[WIP] Chain replace cli commands

dingo-d:
I noticed that we are always doing things step by step by step. And I thought to myself: 

_Self, wouldn't it be cooler if you could just chain these methods? Like the cool devs in Laravel and Symfony are doing?_

Sure enough, turns that you can do that.

> But why would you do that??

Because I can 😆  (evil laugh lol).

Not sure about the performance gain, but it just looks better.

```php
public function __invoke(array $args, array $assocArgs) // phpcs:ignore
{
    $className = $this->getClassShortName();

    // Read the template contents and replace the placeholders with provided variables.
    $class = $this->getExampleTemplate(__DIR__, $className);

    // Replace stuff in file.
    $class = $this->renameClassName($className, $class);

    $class = $this->renameNamespace($assocArgs, $class);

    $class = $this->renameUse($assocArgs, $class);

    // Output final class to new file/folder and finish.
    $this->outputWrite(static::OUTPUT_DIR, $className, $class, $assocArgs);
}
```

vs 

```php
public function __invoke(array $args, array $assocArgs) // phpcs:ignore
{
    $className = $this->getClassShortName();

    // Read the template contents and replace the placeholders with provided variables.
    $this->getExampleTemplate(__DIR__, $className)
        ->renameClassName($className)
        ->renameNamespace($assocArgs)
        ->renameUse($assocArgs)
        ->outputWrite(static::OUTPUT_DIR, $className, $assocArgs);
}
```

The commands just flow one after another, doing their thing. Beautiful. Clean. Efficient.

Ok, not 100% sure about the efficiency, but when I ran this command it was darn fast. Not sure if that was the rush of this newfound power or something to do with the fact I'm no longer overwriting the variable all the time and passing it around, and instead just writing to a member variable. I'd have to profile these calls and check (I wrote it as a to do in my little book).

**This is still a WIP**. All the other renaming CLI classes need to be changed and maybe some other helper methods.

This PR should be reviewed by commits. I anticipate it will be a bit larger, as every CLI class will have to be changed, so it's going to be easier to review it one commit at a time.

Also, I've put most of the commands back to the abstract class. No real need for them to be in the trait, as they are not reused anywhere 🤷‍♂️  Except the error one, that's actually used all over the place.


IvanGrginovInf (approved):

iobrado (approved):

iruzevic (changes requested):
Looks awesome, but we are not going to merge this in an release/5.0.0. This needs more testing and more changes on all of the files. Release 5 is tested (with a few bugs but it works), we don't have the time to test everything again. 

I have created a new branch called release/5.0.1 so please create a PR on that branch and we will merge this in an release after this one.

IvanGrginovInf (approved):

iruzevic (approved):

PR #108, opened 2020-12-04
Chain replace cli commands

dingo-d:
I noticed that we are always doing things step by step by step. And I thought to myself: 

_Self, wouldn't it be cooler if you could just chain these methods? Like the cool devs in Laravel and Symfony are doing?_

Sure enough, turns that you can do that.

> But why would you do that??

Because I can 😆  (evil laugh lol).

Not sure about the performance gain, but it just looks better.

```php
public function __invoke(array $args, array $assocArgs) // phpcs:ignore
{
    $className = $this->getClassShortName();

    // Read the template contents and replace the placeholders with provided variables.
    $class = $this->getExampleTemplate(__DIR__, $className);

    // Replace stuff in file.
    $class = $this->renameClassName($className, $class);

    $class = $this->renameNamespace($assocArgs, $class);

    $class = $this->renameUse($assocArgs, $class);

    // Output final class to new file/folder and finish.
    $this->outputWrite(static::OUTPUT_DIR, $className, $class, $assocArgs);
}
```

vs 

```php
public function __invoke(array $args, array $assocArgs) // phpcs:ignore
{
    $className = $this->getClassShortName();

    // Read the template contents and replace the placeholders with provided variables.
    $this->getExampleTemplate(__DIR__, $className)
        ->renameClassName($className)
        ->renameNamespace($assocArgs)
        ->renameUse($assocArgs)
        ->outputWrite(static::OUTPUT_DIR, $className, $assocArgs);
}
```

The commands just flow one after another, doing their thing. Beautiful. Clean. Efficient.

Ok, not 100% sure about the efficiency, but when I ran this command it was darn fast. Not sure if that was the rush of this newfound power or something to do with the fact I'm no longer overwriting the variable all the time and passing it around, and instead just writing to a member variable. I'd have to profile these calls and check (I wrote it as a to do in my little book).

**This is still a WIP**. All the other renaming CLI classes need to be changed and maybe some other helper methods.

This PR should be reviewed by commits. I anticipate it will be a bit larger, as every CLI class will have to be changed, so it's going to be easier to review it one commit at a time.

Also, I've put most of the commands back to the abstract class. No real need for them to be in the trait, as they are not reused anywhere 🤷‍♂️  Except the error one, that's actually used all over the place.


iruzevic (approved):

src/Cli/AbstractCli.php
```diff
@@ -183,4 +193,735 @@ public function getCommandName(): string
...
+	 *
+	 * @return string
+	 */
+	public function getFileName(string $fileName): string
+	{
+		if (strpos($fileName, ' ') !== false) {
+			$fileName = strtolower($fileName);
+		}
+
+		$class = explode('_', str_replace('-', '_', str_replace(' ', '_', strtolower($fileName))));
```
iruzevic, line 212: there has been some changes on this files. can you please check this PR and implement the changes to your branch ?

https://github.com/infinum/eightshift-libs/pull/106/files

iruzevic (changes requested):

IvanGrginovInf (approved):

src/Cli/AbstractCli.php
dingo-d: Hmmm, if the PR was merged to branch 3.0.0 I should have picked those changes up because I rebased. Unless you made some changes 10 min ago 😂 

Will look at the diff.

dingo-d: Ok, missed this because this was changed in helpers, not the abstract class. 👍

iruzevic (approved):

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -0,0 +1,302 @@
...
+test('Asserts that selectorCustom returns the correct class attributes are set', function () {
+	$selector = Components::selectorCustom(true, 'button', 'icon', 'blue');
+	$selectorFalse = Components::selectorCustom(false, 'button', 'icon', 'blue');
+
+	$this->assertIsString($selector);
+	$this->assertEquals('button__icon--blue', $selector);
+	$this->assertEquals('', $selectorFalse);
+});
+
+
```
volfkarlo, line 283: I assume this many empty lines are here because of WIP.

volfkarlo (approved):

tests/Helpers/ComponentHelpersTest.php
dingo-d: Good catch!

iruzevic (approved):

PR #110, opened 2020-12-08
Acf theme options fix

kmaric69:
There was an error while adding the Theme Options class.
The error was caused because of the wrong `THEME_OPTIONS_CAPABILITY` and there was no value inside `acf_add_local_field_group`.

I've changed the capability and added an example placeholder field.

### DIFF
--- a/src/ThemeOptions/ThemeOptionsExample.php
+++ b/src/ThemeOptions/ThemeOptionsExample.php
@@ -74,7 +74,30 @@ class ThemeOptionsExample implements ServiceInterface
 	public function registerThemeOptions(): void
 	{
 		if (function_exists('acf_add_options_page')) {
-			acf_add_local_field_group();
+			acf_add_local_field_group(
+				array(
+					'key' => 'group_5fcab51c7138c',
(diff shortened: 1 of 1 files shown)


IvanGrginovInf (approved):

src/ThemeOptions/ThemeOptionsExample.php
```diff
@@ -74,7 +74,30 @@ public function createThemeOptionsPage(): void
 	public function registerThemeOptions(): void
 	{
 		if (function_exists('acf_add_options_page')) {
-			acf_add_local_field_group();
+			acf_add_local_field_group(
+				array(
+					'key' => 'group_5fcab51c7138c',
+					'title' => 'Theme Options',
```
gabriel-glo, line 80: This could be wrapped with translation function.

kmaric69: good catch!

gabriel-glo (approved):

dingo-d (approved):

iruzevic (approved):

PR #122, opened 2020-12-16
Update form allowed html list

gabriel-glo:
Allow `selected` attribute to the `option` element
Allow `checked` and `disabled` attribute to the `input` element


iruzevic (approved):

kmaric69 (approved):

dingo-d (approved):

dingo-d:
`meta` tag should also be added, with the options, so that the head component isn't stripped away (that breaks the responsive)

dingo-d (changes requested):
meta tag should be added

iruzevic (approved):

PR #124, opened 2021-01-06
Fix WP-CLI's autoload mess

szepeviktor:
> Error: Found 2 errors

:)


szepeviktor:
> Error: Found 1 error

dingo-d:
There is one issue that phpstan is showing still. Do you know what we need to do to fix it? 

Also, the unit tests will fail on this branch, which is expected. The release/3.0.1 branch should have updated tests (and tons of fixes in the CLI methods).

dingo-d (approved):

iruzevic (approved):

szepeviktor:
> There is one issue that phpstan is showing still. Do you know what we need to do to fix it?

Yes, that was #125

PR #125, opened 2021-01-06
Improve array processing in AbstractBlocks

szepeviktor:
Fix last PHPStan error.
Please see #124

### DIFF
--- a/src/Blocks/AbstractBlocks.php
+++ b/src/Blocks/AbstractBlocks.php
@@ -438,7 +438,7 @@ abstract class AbstractBlocks implements ServiceInterface, RenderableBlockInterf
 
 		return array_reduce(
 			array_keys($componentAttributes),
-			function ($output, $name) use ($realComponentName, $newComponentName, $componentAttributes) {
+			static function ($output, $name) use ($realComponentName, $newComponentName, $componentAttributes) {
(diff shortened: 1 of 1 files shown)


dingo-d (dismissed):
LGTM 👍🏼  Will have to test this on a project to be 100% sure it works as intended. Thanks!

src/Blocks/AbstractBlocks.php
```diff
@@ -430,20 +430,20 @@ protected function getAttributes(array $blockDetails): array
...
-		} else {
-			$output = $componentAttributes;
+		if ($realComponentName === $newComponentName) {
+			return $componentAttributes;
 		}
 
-		return $output;
+		return array_reduce(
+			array_keys($componentAttributes),
+			function ($output, $name) use ($realComponentName, $newComponentName, $componentAttributes) {
```
szepeviktor, line 441: `static function` is a bit better, faster - I think `"php": "^7.2"` allows it.

dingo-d: Yeah, I think that PhpStorm suggests using `static` inside closures, so you can add it 👍🏼

szepeviktor: Done.

dingo-d (approved):

iruzevic (approved):

Issue #121, opened 2020-12-16
[FEATURE] Write all the tests! 🙆🏼‍♂️


dingo-d:
As discussed on the standup we need to divide and conquer the task of writing the unit tests.

The initial setup and test examples have been merged to the `release/3.1.0` branch so you can branch from there. The GH Actions are also set up, so that will work as well 🙂 .

@infinum/wordpress-team you can comment which parts you'll like to test so that we don't overlap.

gabriel-glo:
- [ ] Config
- [ ] ThemeOptions
- [ ] View
- [ ]  CustomTaxonomy

dingo-d:
- [ ] Helpers
- [ ] CustomPostType
- [ ] CustomMeta
- [ ] Db

iruzevic:
- [x] ModifyAdminAppearance
- [x] Login
- [x] CiExclude

IvanGrginovInf:
[ ] Rest fields
[ ] Rest routes

kmaric69:
For starters:
- [ ] Menu
- [ ] Media

goranalkovic-infinum:
Let's try this out:

- [ ] Manifest

volfkarlo:
I'll try this ones
- [ ] Enqueue/Blocks

iobrado:
I'll try this
 [ ] Enqueue/Theme

IvanGrginovInf:
- [ ] Autowiring

PR #127, opened 2021-01-06
Add additional unit tests

dingo-d:
This is a continuation of #121 where all the tests which I've chosen will be made.

- [x] Helpers
- [x] CustomPostType
- [x] CustomMeta
- [x] Db

Note: Db tests are only partially done, I'll leave this to somebody else who wants to play around with it. I'm drawing a blank on how to test this actually tbh 🤷🏼‍♂️


tests/Helpers.php
```diff
@@ -14,6 +14,17 @@
 // Mock escaping function.
 Functions\when('wp_kses_post')->returnArg();
 
+// Mock json success and error handlers.
+Functions\when('wp_send_json_success')->alias(function(...$args) {
+    echo json_encode($args);
+});
```
IvanGrginovInf, line 20:
```suggestion
Functions\when('wp_send_json_success')->alias(function($data = null, $status_code = null, $options = 0) {
    $response = array( 'success' => true );
    
    if ( isset( $data ) ) {
        $response['data'] = $data;
    }
    
    echo json_encode($response, $options);
});
```

`wp_send_json_success` should add `success => true` to the response. Also I know this is simplified mock but it doesn't hurt to be true to the source, what do you think? https://developer.wordpress.org/reference/functions/wp_send_json_success/

tests/Helpers.php
```diff
@@ -14,6 +14,17 @@
...
+Functions\when('wp_send_json_error')->alias(function(...$args) {
+    echo json_encode($args);
+});
```
IvanGrginovInf, line 23:
```suggestion
Functions\when('wp_send_json_error')->alias(function(...$args) {
    $response = array( 'success' => false );
 
    if ( isset( $data ) ) {
        if ( is_wp_error( $data ) ) {
            $result = array();
            foreach ( $data->errors as $code => $messages ) {
                foreach ( $messages as $message ) {
                    $result[] = array(
                        'code'    => $code,
                        'message' => $message,
                    );
                }
            }
 
            $response['data'] = $result;
        } else {
            $response['data'] = $data;
        }
    }
    
    echo json_encode($response, $options);
});
```

Same as above: https://developer.wordpress.org/reference/functions/wp_send_json_error/. What do you think? It's unlikely these functions will change much and now they behave more correctly.

tests/Helpers/ErrorLoggerTraitTest.php
```diff
@@ -0,0 +1,28 @@
...
+    use ErrorLoggerTrait;
+}
+
+test('Test restResponseHandler will return expected result in case of success', function () {
+    $mockLogger = new MockErrorLogger();
+
+    $response = $mockLogger->restResponseHandler(200, 'Some status', 'Some message', ['fake data']);
+
+    $this->assertJson($response, 'Response is not a JSON string.');
+    $this->assertEquals('[{"code":200,"status":"Some status","message":"Some message","data":["fake data"]},200]', $response);
```
IvanGrginovInf, line 17: If the above changes are implemented then it would be something like this:

```suggestion
    $this->assertEquals('[{"success": "true", "code":200,"status":"Some status","message":"Some message","data":["fake data"]},200]', $response);
```

IvanGrginovInf (commented):
Left a suggestion, not sure if it's needed, let me know what you think, otherwise looks good! @dingo-d

tests/Helpers.php
dingo-d: Sure thing, I'm fine with it. I opted for a simplified version because I'm not really interested in what is going on under the hood of the `wp_send_json_success` function (at the end it's just `echo` the `json_encode` and `die` with some custom headers set). The unit test is only concerned that the method we implemented works as it should.

But I guess this mock cannot hurt 👍🏼  I'll update it and the tests accordingly.

tests/Helpers.php
dingo-d: Fine by me, but, as mentioned above, I am not testing those functions per se. Using real functions would make these integration and not unit tests. That's why I kept them simple.

tests/Helpers.php
IvanGrginovInf: Yeah I 100% understand. Only mentioned this because I could imagine a scenario where I write a test for an arbitrary route and assert that `success` key is set to ensure it was called successfully (as you'd test this in your app code).

IvanGrginovInf (approved):

iruzevic (approved):

dingo-d:
I have some issues with mocking and checking the post type examples with Brain Monkey and Pest atm. Hopefully, I'll get some info on how to resolve it

https://github.com/Brain-WP/BrainMonkey/issues/92
https://github.com/pestphp/pest/issues/259

tests/CustomPostType/CustomPostTypeExampleTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Tests\Unit\CustomPostType;
+
+use Brain\Monkey;
+use EightshiftBoilerplate\CustomPostType\PostTypeExample;
+
+/**
+ * Mock before tests.
+ */
+//beforeEach(function () {
```
iruzevic, line 11: come commented methods here?

tests/coverage/clover.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<coverage generated="1610609739">
-  <project timestamp="1610609739">
+<coverage generated="1610819159">
```
iruzevic, line 2: does this generally go in the repo?

IvanGrginovInf (approved):

tests/CustomPostType/CustomPostTypeExampleTest.php
dingo-d: Yeah, I commented this out so that the tests don't fail. I'm waiting for some guidance from the pest/brain monkey for this.

tests/coverage/clover.xml
dingo-d: I searched a bit and didn't find any agreement on this tbh 🤷🏼‍♂️  I thought about removing it, since we are generating coverage in the GH action anyhow, so in theory, I can remove it.

iruzevic (approved):

dingo-d:
@infinum/wordpress-team I've added additional tests. The only thing that is a bit harder to test is the DB part. So I'll leave this to anybody who's interested 😄 

I've solved the issues with mocks from Brain Monkey deleting my helpers by wrapping them in a function and calling that function inside the `beforeEach()` function.

The best way to understand my train of thought is to go through the commit messages. there is some explanation there what I did, why, and how I did it.

Basically, all other CLI tests can now be emulated like the ones here for the CPT and menu one.

goranalkovic-infinum (approved):

tests/CustomPostType/LabelGeneratorTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+	$this->generator = new LabelGenerator();
+});
+
+
+test('Generating labels will work', function () {
+	$nouns = [
+		LabelGenerator::SINGULAR_NAME_UC => 'Book',
+		LabelGenerator::SINGULAR_NAME_LC => 'book',
+        LabelGenerator::PLURAL_NAME_UC => 'Books',
```
IvanGrginovInf, line 22:
```suggestion
		LabelGenerator::PLURAL_NAME_UC => 'Books',
```

tests/CustomPostType/LabelGeneratorTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+	$this->assertSame($expectedArray, $generatedLabels);
+});
+
+
+test('Generating labels will throw exception if noun wasn\'t specified', function () {
+	$nouns = [
+		LabelGenerator::SINGULAR_NAME_UC => 'Book',
+		LabelGenerator::SINGULAR_NAME_LC => 'book',
+        LabelGenerator::PLURAL_NAME_UC => 'Books',
```
IvanGrginovInf, line 66:
```suggestion
		LabelGenerator::PLURAL_NAME_UC => 'Books',
```

IvanGrginovInf (approved):
Left 2 comments regarding spacing (tabs, not spaces) otherwise looks good!

PR #131, opened 2021-01-12
updating env handler

iruzevic:
- switching to native ways for env handler
- removing never used methods because we are using imposter and we don't need to prefix our stuff

### DIFF
--- a/src/Config/AbstractConfigData.php
+++ b/src/Config/AbstractConfigData.php
@@ -19,24 +19,6 @@ abstract class AbstractConfigData implements ConfigDataInterface
 {
 
 	/**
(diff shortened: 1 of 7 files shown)


IvanGrginovInf (approved):

dingo-d:
I'm a bit confused. What about the `release/3.0.1` branch, with updated commands, and unit tests? I thought this will be the next release point. Shouldn't we either merge that to develop or branch from that branch?

Because the changes you've made here were/are covered by tests, so if you changed something you'll see how it will affect the tests, and ensure nothing gets badly broken.

iruzevic:
@dingo-d Lets discus this on internal channels

dingo-d (approved):

PR #135, opened 2021-01-13
Update ci.yml

szepeviktor:
Just playing...


szepeviktor:
This is very bad.
Techdebt goes under the carpet.

dingo-d:
I've added the phpstan to the release branch right now, so we should pick those up and solve them accordingly, don't worry 😄

PR #141, opened 2021-02-05
Remove debugging if not development setup

kmaric69:
I think we should not have debugging turned on when in production or staging.

### DIFF
--- a/src/ConfigProject/ConfigProjectExample.php
+++ b/src/ConfigProject/ConfigProjectExample.php
@@ -32,12 +32,14 @@ define('AUTOSAVE_INTERVAL', 240);
 // Disable automatic updating of plugins.
 define('AUTOMATIC_UPDATER_DISABLED', true);
 
-// Environment based setup.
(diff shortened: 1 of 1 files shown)


dingo-d (approved):

IvanGrginovInf:
Not sure I agree. First of all it's easier to work with locally (even if somebody doesn't use the WP env). Second, I'd leave debugging on staging, sometimes it's needed and it's 99% fine as stagings are private (behind basic auth / firewall)

dingo-d:
Then we'd need a separate if for staging, prod, and development. Which I'm also fine with 🙂 

We just need to prevent logging on production.

iruzevic (dismissed):

kmaric69:
I wrapped it inside a condition for production specific. This solves our problem.

dingo-d (approved):

iruzevic (approved):

gabriel-glo (approved):

IvanGrginovInf (approved):

PR #143, opened 2021-02-20
WIP Update bem menu

dingo-d:
A rewrite of the old BEM menu. Relates to changes in https://github.com/infinum/eightshift-frontend-libs/pull/292


IvanGrginovInf (approved):

kmaric69 (approved):

dingo-d:
Blocked by https://github.com/infinum/eightshift-libs/issues/144 for testing. Once fixed will test this out on a real project and merge.

iruzevic (approved):

dingo-d:
I just realized that by modifying the parameters of the `bemMenu` method I'm causing a breaking change. Not only that but during testing, I realized that we have a tight coupling with the menu walker, which should be added as a dependency for easier testing. I'll put this to a later release date.

IvanGrginovInf (approved):

iruzevic:
@dingo-d What is up with this pull? Is this ready?

dingo-d:
@iruzevic technically yes, but it will be a big bc break, so if you update the libs on an older project you could have certain things break in the project.

We can discuss this on the standup.

PR #145, opened 2021-02-23
Feature/render helper

iruzevic:
I have found out that when you want to use component in a different component the is smart, for example use heading component for single article, you have to provide attribute for the headingContent. You expect to have all default attributes from the component but this is not happening out of the box. So you must map all the default components manually.

Now you can provide 4th key as `true` and this will happen automatically. 

Before:
```php
echo \wp_kses_post(
	Components::render(
		'card-review-alternative',
		[
			'headingColor' => $cardReviewAlternativeManifest['attributes']['headingColor']['default'],
			'headingSize' => $cardReviewAlternativeManifest['attributes']['headingSize']['default'],
			'headingWeight' => $cardReviewAlternativeManifest['attributes']['headingWeight']['default'],
			/* translators: %s will be replaced with the number of rating. */
			'headingContent' => sprintf(__('Rated %s/5', 'Tvornica'), $rating),

			'ratingAmount' => $rating,

			'paragraphContent' => \get_field(ReviewAcfMeta::FIELD_REVIEW_KEY, $review->ID),
			'paragraphSize' => $cardReviewAlternativeManifest['attributes']['paragraphSize']['default'],

			'linkContent' => __('Read the reviews', 'Tvornica'),
			'linkUrl' => "#reviews",
			'linkSize' => $cardReviewAlternativeManifest['attributes']['linkSize']['default'],
			'linkColor' => $cardReviewAlternativeManifest['attributes']['linkColor']['default'],
			'linkIsUppercase' => $cardReviewAlternativeManifest['attributes']['linkIsUppercase']['default'],
			'linkUseArrow' => $cardReviewAlternativeManifest['attributes']['linkUseArrow']['default'],

			'blockClass' => $componentClass
		],
	)
);
```

now:
```php
echo \wp_kses_post(
	Components::render(
		'card-review-alternative',
		[
			/* translators: %s will be replaced with the number of rating. */
			'headingContent' => sprintf(__('Rated %s/5', 'Tvornica'), $rating),

			'ratingAmount' => $rating,

			'paragraphContent' => \get_field(ReviewAcfMeta::FIELD_REVIEW_KEY, $review->ID),

			'linkContent' => __('Read the reviews', 'Tvornica'),
			'linkUrl' => "#reviews",

			'blockClass' => $componentClass
		],
	),
	'',
	true
);
```

All the default components will be inherited from the called component

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -210,7 +210,7 @@ class Components
 			$manifestKey = $manifest['attributes'][$key] ?? null;
 
 			if ($manifestKey === null) {
-				throw new \Exception("{$key} key does not exist in the {$componentName} component. Please check your implementation. Check if your {$key} attributes exists in the components manifest.json");
+				throw new \Exception("{$key} key does not exist in the {$componentName} component. Please check your implementation. Check if your {$key} attribut exists in the component's manifest.json");
 			}
 
 			$defaultType = $manifestKey['type'];
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -74,13 +74,13 @@ class Components
 	 * @param array  $attributes Array of attributes that's implicitly passed to component.
 	 * @param string $parentPath If parent path is provides it will be appended to the file location.
 	 *                            If not get_template_directory_uri() will be used as a default parent path.
-	 * @param bool   $useParentDefaults If true the helper will fetch component manifest and merge default attributes in the original attributes list.
+	 * @param bool   $useComponentDefaults If true the helper will fetch component manifest and merge default attributes in the original attributes list.
 	 *
 	 * @throws \Exception When we're unable to find the component by $component.
 	 *
 	 * @return string
 	 */
-	public static function render(string $component, array $attributes = [], string $parentPath = '', bool $useParentDefaults = false)
+	public static function render(string $component, array $attributes = [], string $parentPath = '', bool $useComponentDefaults = false)
 	{
 		if (empty($parentPath)) {
 			$parentPath = \get_template_directory();
@@ -90,13 +90,13 @@ class Components
 		if (strpos($component, '.php') !== false) {
 			$componentPath = "{$parentPath}/$component";
 
-			if ($useParentDefaults) {
+			if ($useComponentDefaults) {
 				$manifest = self::getManifest($parentPath);
 			}
 		} else {
 			$componentPath = "{$parentPath}/src/Blocks/components/{$component}/{$component}.php";
 
-			if ($useParentDefaults) {
+			if ($useComponentDefaults) {
 				$manifest = self::getManifest("{$parentPath}/src/Blocks/components/{$component}");
 			}
 		}
@@ -105,7 +105,7 @@ class Components
 			ComponentException::throwUnableToLocateComponent($componentPath);
 		}
 
-		if ($useParentDefaults && isset($manifest['attributes'])) {
+		if ($useComponentDefaults && isset($manifest['attributes'])) {
 			$defaultAttributes = [];
 			foreach ($manifest['attributes'] as $itemKey => $itemValue) {
 				if (isset($itemValue['default'])) {
@@ -121,7 +121,7 @@ class Components
 		// Wrap component with parent BEM selector if parent's class is provided. Used
 		// for setting specific styles for components rendered inside other components.
 		if (isset($attributes['parentClass'])) {
-			echo "<div class=\"{$attributes['parentClass']}__{$component}\">"; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			printf('<div class="%s">', \esc_attr("{$attributes['parentClass']}__{$component}"));
 		}
 
 		require $componentPath;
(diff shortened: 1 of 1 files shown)


IvanGrginovInf:
Hmm, shouldn't this already happen by default? Isn't that what `checkAttr` is for? (i.e. use the correct attribute based on `componentName` if passed, otherwise default to component's manifest?)

src/Helpers/Components.php
```diff
@@ -74,12 +74,13 @@ public static function classnames(array $classes): string
...
 	 * @param string $parentPath If parent path is provides it will be appended to the file location.
 	 *                            If not get_template_directory_uri() will be used as a default parent path.
+	 * @param bool   $useParentDefaults If true the helper will fetch component manifest and merge default attributes in the original attributes list.
 	 *
 	 * @throws \Exception When we're unable to find the component by $component.
 	 *
 	 * @return string
 	 */
-	public static function render(string $component, array $attributes = [], string $parentPath = '')
+	public static function render(string $component, array $attributes = [], string $parentPath = '', bool $useParentDefaults = false)
```
IvanGrginovInf, line 83: This should be called `useComponentDefaults`,  (or something without `parent` in it), no? There's no parent here, you're rendering a component and reading it's manifest.

src/Helpers/Components.php
```diff
@@ -88,20 +89,39 @@ public static function render(string $component, array $attributes = [], string
 		// Detect if user passed component name or path.
 		if (strpos($component, '.php') !== false) {
 			$componentPath = "{$parentPath}/$component";
+
+			if ($useParentDefaults) {
+				$manifest = self::getManifest($parentPath);
```
IvanGrginovInf, line 94: There's seemingly no `getManifest()` method in this class.

src/Helpers/Components.php
```diff
@@ -88,20 +89,39 @@ public static function render(string $component, array $attributes = [], string
...
+			$attributes = array_merge($defaultAttributes, $attributes);
+		}
+
 		ob_start();
 
 		// Wrap component with parent BEM selector if parent's class is provided. Used
 		// for setting specific styles for components rendered inside other components.
 		if (isset($attributes['parentClass'])) {
-			echo \wp_kses_post("<div class=\"{$attributes['parentClass']}__{$component}\">");
+			echo "<div class=\"{$attributes['parentClass']}__{$component}\">"; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
IvanGrginovInf, line 124: This should still be escaped, you can use `printf` for this:

```php
<?php

printf('<div class="%s">', esc_attr( "$attributes['parentClass']}__{$component}" ) );
```

IvanGrginovInf (changes requested):
Left a few comments

src/Helpers/Components.php
iruzevic: ?? not sure what the question is but getManifest is a few lines down

iruzevic:
> Hmm, shouldn't this already happen by default? Isn't that what `checkAttr` is for? (i.e. use the correct attribute based on `componentName` if passed, otherwise default to component's manifest?)

checkAttr does this but only if you are using "native" way of calling component using the manifest.json components key.
My use case is that I need to call a component that is not editable from the admin but is populated manually. 
Here is an example https://github.com/infinum/wordpress-tvornica-snova/blob/staging/wp-content/themes/tvornica/index.php

src/Helpers/Components.php
dingo-d: I think the confusion comes from the fact that GH hides most of the file during review and Grga didn't notice the method on line 145 🙂

src/Helpers/Components.php
```diff
@@ -190,7 +210,7 @@ public static function checkAttr(string $key, array $attributes, array $manifest
 			$manifestKey = $manifest['attributes'][$key] ?? null;
 
 			if ($manifestKey === null) {
-				throw new \Exception("{$key} key does not exist in the {$componentName} component. Please check your implementation.");
+				throw new \Exception("{$key} key does not exist in the {$componentName} component. Please check your implementation. Check if your {$key} attributes exists in the components manifest.json");
```
dingo-d, line 213:
```suggestion
				throw new \Exception("{$key} key does not exist in the {$componentName} component. Please check your implementation. Check if your {$key} attribut exists in the component's manifest.json");
```

src/Helpers/Components.php
iruzevic: updated

dingo-d (changes requested):
Left one minor change. We need to agree on what to do with the 3.1.0 branch. As in, when do we want to release it, because if this feature is merged, that branch will have to be rebased and tests written to accommodate this change.

iruzevic:
> Left one minor change. We need to agree on what to do with the 3.1.0 branch. As in, when do we want to release it, because if this feature is merged, that branch will have to be rebased and tests written to accommodate this change.

I have updated the fix. Lets agree today about 3.1.0

dingo-d (approved):

IvanGrginovInf (approved):

src/Helpers/Components.php
IvanGrginovInf: Ah my bad, thought I expanded the entire file 🙈

PR #148, opened 2021-02-28
Fix CI

szepeviktor:
Codeception does not want to start.
What to do?


szepeviktor:
It was dropped :(
https://github.com/infinum/eightshift-libs/tree/release/3.1.0

PR #149, opened 2021-02-28
Improve CI

szepeviktor:
- run only on pushes/PR-s to develop branch
- YAML strings
- execute `composer update`
- ~~explicit Composer options~~ got a 👎

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - develop
-      - master
+      - main
     paths-ignore:
       - '**.md'
   pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - develop
+      - master
     paths-ignore:
       - '**.md'
   pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Check coding standards using PHPCS
         continue-on-error: true
-        run: composer run-script standards:check -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
+        run: composer standards:check -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
 
       - name: Show PHPCS results in PR
         run: cs2pr ./phpcs-report.xml
@@ -78,4 +78,4 @@ jobs:
          dependency-versions: highest
 
      - name: Check code consistency using PHPStan
-       run: composer run-script analyze
+       run: composer analyze
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,13 @@ name: Continuous integration checks
 on:
   # Run on PRs and pushes, only on significant changes.
   push:
+    branches:
+      - develop
     paths-ignore:
       - '**.md'
   pull_request:
+    branches:
+      - develop
     paths-ignore:
       - '**.md'
 
@@ -16,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.3, 7.4]
+        php: ['7.3', '7.4']
 
     steps:
       # Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,9 +3,13 @@ name: Continuous integration checks
 on:
   # Run on PRs and pushes, only on significant changes.
   push:
+    branches:
+      - develop
```
dingo-d, line 7: Why only on `develop`? Releases are done from the `main` branch (currently `master` branch).

.github/workflows/ci.yml
```diff
@@ -34,11 +38,13 @@ jobs:
...
       - name: Install Composer dependencies
-        uses: "ramsey/composer-install@v1"
+        uses: ramsey/composer-install@v1
+        with:
+          dependency-versions: highest
 
       - name: Check coding standards using PHPCS
         continue-on-error: true
-        run: composer standards:check -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
+        run: composer run-script standards:check -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
```
dingo-d, line 47: `run-script` is only necessary for standard composer scripts for specific events (post-update, post-install, etc.). So here it can be omitted 🙂

.github/workflows/ci.yml
```diff
@@ -66,7 +72,9 @@ jobs:
...
      # @link https://github.com/marketplace/actions/install-composer-dependencies
      - name: Install Composer dependencies
-       uses: "ramsey/composer-install@v1"
+       uses: ramsey/composer-install@v1
+       with:
+         dependency-versions: highest
 
      - name: Check code consistency using PHPStan
-       run: composer analyze
+       run: composer run-script analyze
```
dingo-d, line 80: Same as the above comment. You can omit this.

.github/workflows/test.yml
```diff
@@ -3,9 +3,13 @@ name: Unit tests
 on:
   # Run on PRs and pushes, only on significant changes.
   push:
+    branches:
+      - develop
```
dingo-d, line 7: The same question as in the previous workflow.

.github/workflows/test.yml
```diff
@@ -20,19 +24,21 @@ jobs:
...
+        with:
+          dependency-versions: highest
 
       - name: Run unit tests
-        run: composer test:unit
+        run: composer run-script test:unit
 
       - name: Run coverage generation
-        run: composer test:coverage -q # We don't need an output for coverage, codecov will do that.
+        run: composer run-script test:coverage -q # We don't need an output for coverage, codecov will do that.
```
dingo-d, line 41: `run-script` can be omitted.

dingo-d (changes requested):
Left some questions and minor change requests. Thanks for the contribution and cleanup 👍🏼

.github/workflows/ci.yml
szepeviktor: Added.

.github/workflows/ci.yml
szepeviktor: `standards:check` is the name of a Composer script, it is not a Composer command.
`composer run-script` is lazy.
This is explicit programing.

.github/workflows/test.yml
szepeviktor: Added.

.github/workflows/ci.yml
szepeviktor: `composer run-script --list` vs. `composer list`

szepeviktor: > `run-script` is only necessary for standard composer scripts for specific events (post-update, post-install, etc.).

This may not be true. Try `composer post-update` ...

.github/workflows/ci.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - develop
+      - master
```
dingo-d, line 8:
```suggestion
      - main
```

.github/workflows/ci.yml
dingo-d: Don't think there is `post-update`, you mean `post-update-cmd`? This is a standard composer event and as such requires a `run-script` to run it.

Scripts that are defined in the `scripts` part of the `composer.json` such as `standards:check` do not require `run-script` in front of it and will work without it. I don't think we gain anything by adding it to the workflow command.

.github/workflows/test.yml
```diff
@@ -3,9 +3,14 @@ name: Unit tests
 on:
   # Run on PRs and pushes, only on significant changes.
   push:
+    branches:
+      - develop
+      - master
```
dingo-d, line 8:
```suggestion
      - main
```

My colleagu changed the `master` to `main` branch in the meantime 🙈

dingo-d (changes requested):
Comments left and some small changes

.github/workflows/ci.yml
szepeviktor: Okay I see.
I rather run slow 🚶‍♂️ than be implicit.

szepeviktor:
Please squash my commits.

dingo-d (approved):

IvanGrginovInf (approved):

szepeviktor:
kép

dingo-d:
I'll merge this ASAP, I just need to get my colleague to enable squashing on the PRs 😄

PR #155, opened 2021-03-09
adding fix to helper

iruzevic:
- A bunch of helpers to implement css variables.
- Refactored render method for easier test
- Added multiple methods for checking variables
- Tests for the whole Component class


tests/data/src/Blocks/components/button/manifest.json
```diff
@@ -27,19 +27,24 @@
     },
     "buttonSize": {
       "type": "string",
-      "default": "default"
+      "default": "default",
```
goranalkovic-infinum, line 30: `defaults` means "use global", right? If that's the case, could we use `null` as "use global" and make manifests leaner?

tests/data/src/Blocks/components/button/manifest.json
```diff
@@ -27,19 +27,24 @@
...
+      "variable": true
     },
     "buttonColor": {
       "type": "string",
-      "default": "primary"
+      "default": "primary",
+      "variable": true,
```
goranalkovic-infinum, line 36: Could `variable` be `true` by default and then if it needs to be `false` it's easy to override?

tests/data/src/Blocks/manifest.json
```diff
@@ -0,0 +1,69 @@
...
+				"slug": "light",
+				"color": "#CCCCCC"
+			},
+			{
+				"name": "White",
+				"slug": "white",
+				"color": "#FFFFFF"
+			}
+		],
+		"gradients": [
```
goranalkovic-infinum, line 54: What's the difference between `gradients` and `colors` if both contain only a single swatch?

Could we have gradients as:
```json
{
    "name": "Black to transparent",
    "slug": "black-to-transparent",
    "stops": [
        "#000000 0%",
        "transparent 100%"
    ]
```

IvanGrginovInf (approved):

tests/Helpers.php
```diff
@@ -21,6 +21,9 @@ function setupMocks() {
 	// Mock escaping function.
 	Functions\when('wp_kses_post')->returnArg();
 
+	// Mock rand function.
+	Functions\when('wp_rand')->justReturn(1154790670);
```
dingo-d, line 25: You could have aliased it with `mt_rand` but this is ok for testing purposes I guess 😄

dingo-d (approved):
👍🏼

gabriel-glo (approved):

tests/data/src/Blocks/components/button/manifest.json
iruzevic: no default is value set on the attribute by default. Not sure what you are implying here

tests/data/src/Blocks/components/button/manifest.json
iruzevic: no because in general you would need 2-3 variables from attributes and everything else is ignored

tests/data/src/Blocks/manifest.json
iruzevic: Basically this is everything used from the native core. I don't want to spend any more time on this.

PR #160, opened 2021-03-10
Test CLI part

dingo-d:
Testing CLI commands.

I've simplified the way mockery mock is created with a helper. This caused a lot of files to be changed.

Also, I've fixed one of the block tests which was failing. I need to test this further, because of the changes in the `getExampleTemplate` method.

Screenshot 2021-03-22 at 21 40 02

Current coverage: 58.00% 🎉

### DIFF
--- a/src/Cli/Cli.php
+++ b/src/Cli/Cli.php
@@ -44,14 +44,6 @@ use EightshiftLibs\ThemeOptions\ThemeOptionsCli;
  */
 class Cli
 {
-
-	/**
-	 * Top level commands name.
-	 *
-	 * @var string
-	 */
-	protected $commandParentName;
-
 	/**
 	 * All classes and commands that can be used on development and public WP CLI.
 	 *
@@ -181,29 +173,6 @@ class Cli
 
 					break;
 				}
-			}
-		}
-	}
-
-	/**
-	 * Run all CLI commands for normal WPCLI.
-	 *
-	 * @param string $commandParentName Define top level commands name.
-	 *
-	 * @throws \ReflectionException Exception if the class doesn't exist.
-	 *
-	 * @return void
-	 */
-	public function load(string $commandParentName): void
-	{
-		$this->commandParentName = $commandParentName;
-
-		foreach ($this->getPublicClasses() as $item) {
-			$reflectionClass = new \ReflectionClass($item);
-			$class = $reflectionClass->newInstanceArgs([$this->commandParentName]);
-
-			if ($class instanceof CliInterface) {
-				$class->register();
 			}
 		}
 	}
--- a/src/Cli/Cli.php
+++ b/src/Cli/Cli.php
@@ -176,4 +176,25 @@ class Cli
 			}
 		}
 	}
+
+	/**
+	 * Run all CLI commands for normal WPCLI.
+	 *
+	 * @param string $commandParentName Define top level commands name.
+	 *
+	 * @throws \ReflectionException Exception if the class doesn't exist.
+	 *
+	 * @return void
+	 */
+	public function load(string $commandParentName): void
+	{
+		foreach ($this->getPublicClasses() as $item) {
+			$reflectionClass = new \ReflectionClass($item);
+			$class = $reflectionClass->newInstanceArgs([$commandParentName]);
+
+			if ($class instanceof CliInterface) {
+				$class->register();
+			}
+		}
+	}
 }
--- a/tests/Blocks/BlockCliTest.php
+++ b/tests/Blocks/BlockCliTest.php
@@ -3,14 +3,15 @@
 namespace Tests\Unit\Block;
 
 use EightshiftLibs\Blocks\BlockCli;
-use Brain\Monkey\Functions;
+
 use function Tests\deleteCliOutput;
+use function Tests\mock;
 
 /**
  * Mock before tests.
  */
 beforeEach(function () {
-	$wpCliMock = \Mockery::mock('alias:WP_CLI');
+	$wpCliMock = mock('alias:WP_CLI');
 
 $wpCliMock
 	->shouldReceive('success')
--- a/tests/Blocks/BlockCliTest.php
+++ b/tests/Blocks/BlockCliTest.php
@@ -21,6 +21,10 @@ $wpCliMock
 	->shouldReceive('error')
 	->andReturnArg(0);
 
+$wpCliMock
+	->shouldReceive('log')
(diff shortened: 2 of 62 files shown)


tests/Cli/CliInitProjectTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+
+test('CliInitProject CLI documentation is correct', function () {
+	$documentation = $this->cliInitProject->getDoc();
+
+	$key = 'shortdesc';
+
+	$this->assertIsArray($documentation);
+	$this->assertArrayHasKey($key, $documentation);
+	$this->assertEquals('Generates initial setup for WordPress theme project with all files to run a client project, for example: gitignore file for the full WordPress project, continuous integration exclude files, etc.', $documentation[$key]);
```
IvanGrginovInf, line 64: Maybe it's a bit too much? Because now the test is coupled to the documentation copy being exactly like this. (i.e. if we change a word / fix a typo in that, it will break this test). Or is that maybe the intention? What do you think?

dingo-d: That's the point of the test. If we change the documentation the test must be changed as well.

Tests shouldn't be set in stone. They are a validation. If you never changed the tests, the architecture of your app would suffer. They should evolve with the app 🙂

IvanGrginovInf (approved):

dingo-d:
@infinum/wordpress-team when you have the time feel free to look at the PR. I still have reset, runAll and showAll to test, but since this is a bigger PR (the 7b86f96 commit added lot's of changes when I added a helper). it would be good to check it earlier 🙂

iruzevic (changes requested):

src/Cli/Cli.php
```diff
@@ -184,27 +176,4 @@ public function loadDevelop(array $args = []): void
...
-	/**
-	 * Run all CLI commands for normal WPCLI.
-	 *
-	 * @param string $commandParentName Define top level commands name.
-	 *
-	 * @throws \ReflectionException Exception if the class doesn't exist.
-	 *
-	 * @return void
-	 */
-	public function load(string $commandParentName): void
```
iruzevic, line 197: why is this deleted?

tests/Blocks/BlockCliTest.php
```diff
@@ -32,38 +37,27 @@
 	deleteCliOutput($output);
 });
 
-// test('Block CLI command will correctly copy the Block class with defaults', function () {
```
iruzevic, line 35: there are a lot of commented code here

tests/Cli/AbstractCliTest.php
```diff
@@ -0,0 +1,268 @@
...
+	protected $fileContents = 'use EightshiftBoilerplateVendor\Service; use EightshiftBoilerplate\Test;';
+
+	public function __construct(string $commandParentName)
+	{
+		parent::__construct($commandParentName);
+	}
+
+	public function __invoke(array $args, array $assocArgs)
+	{
+		// TODO: Implement __invoke() method.
```
iruzevic, line 21: to do

tests/Cli/CliTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+	Monkey\tearDown();
+});
+
+
+test('Cli getDevelopClasses return correct class list', function () {
+	$developClasses = $this->cli->getDevelopClasses();
+
+	$this->assertIsArray($developClasses);
+	$this->assertTrue(count($developClasses) === 31, 'Total number of classes is correct');
```
iruzevic, line 70: once we add new featured we should update this number and number on line 86?

iruzevic (changes requested):

tests/Cli/CliResetTest.php
```diff
@@ -0,0 +1,60 @@
...
+test('CliReset works', function () {
+	Functions\when('system')->alias(function ($args) {
+		putenv("SYSTEM_CALLED={$args}");
+	});
+
+	$cliReset = $this->cliReset;
+	$cliReset([], []);
+
+	$this->assertSame('Output directory successfully removed.', getenv('SUCCESS'));
+	$this->assertSame('rm -rf \'/Users/denis.zoljom/Projects/Infinum/Eightshift/eightshift-libs/cliOutput/\'', getenv('SYSTEM_CALLED'));
```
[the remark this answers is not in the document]
dingo-d: Oh, my bad! I'll fix this. Totally didn't register my own file path 😅  Good catch! 👍🏼

src/Cli/Cli.php
dingo-d: This wasn't used anywhere in the lib 🤷🏼‍♂️  What is this used for?

tests/Blocks/BlockCliTest.php
dingo-d: That's your commented code 😂  I just wanted to try testing the above test because one of the methods in the CLI part was only used for blocks copying or something like that so I wanted to see if that functionality would break if I change the CLI method 😄

tests/Cli/AbstractCliTest.php
dingo-d: Will remove. PhpStorm adds the stubs automatically. 👍🏼

tests/Cli/CliTest.php
dingo-d: Yes. If this test fails you need to update the number. I'll add them as a variable so that it's a bit more clear what they denote.

src/Cli/Cli.php
iruzevic: Without this nothing is working:

https://github.com/infinum/eightshift-boilerplate/blob/6c49f0a26e053dd3e99e32e53c07fd0dfac39bbf/functions.php#L47

tests/Blocks/BlockCliTest.php
iruzevic: 😄

src/Cli/Cli.php
dingo-d: Oh it's used in the boilerplate! 🤦‍♂️

I'll return it and test. 👍

iruzevic (approved):

src/Cli/Cli.php
dingo-d: Done

dingo-d:
Should be all ok now. Waiting from others from the team to take a look 🙂

gabriel-glo (approved):

PR #161, opened 2021-03-12
Setup tests

goranalkovic-infinum:
This took way more than I hoped it would, ran into some weird issues and path brainf##ks, but all is good now 😄 

Changelog:
- `AbstractCli` wasn't detecting the test environment correctly, added the condition that makes it work now
- extracted command name to a const in `SetupCli` and `UpdateCli`
- added check in `UpdateCli` to detect if we're in a test environment and make it point to `setup.json` in `cliOutput` instead of libs root folder
- added tests for `SetupCli` and `UpdateCli`

### DIFF
--- a/tests/Setup/SetupCliTest.php
+++ b/tests/Setup/SetupCliTest.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace Tests\Unit\Setup;
+
+use EightshiftLibs\Setup\SetupCli;
+
+use function Tests\deleteCliOutput;
+
+/**
+ * Mock before tests.
+ */
+beforeEach(function () {
+    $wpCliMock = \Mockery::mock('alias:WP_CLI');
+
+	$wpCliMock
+		->shouldReceive('success')
+		->andReturnArg(0);
+
+	$wpCliMock
+		->shouldReceive('error')
+		->andReturnArg(0);
+
+	$this->setup = new SetupCli('boilerplate');
+});
+
+/**
+ * Cleanup after tests.
+ */
+afterEach(function () {
+	$output = dirname(__FILE__, 3) . '/cliOutput';
+
+	deleteCliOutput($output);
+});
+
+test('Setup CLI command will correctly copy the Setup class with defaults', function () {
+	$setup = $this->setup;
+	$setup([], $setup->getDevelopArgs([]));
+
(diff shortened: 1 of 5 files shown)


tests/Setup/SetupCliTest.php
```diff
@@ -0,0 +1,68 @@
...
+	$wpCliMock
+		->shouldReceive('success')
+		->andReturnArg(0);
+
+	$wpCliMock
+		->shouldReceive('error')
+		->andReturnArg(0);
+
+	$this->setup = new SetupCli('boilerplate');
+});
```
dingo-d, line 24: There seem to be some issues with the indentation here.

tests/Setup/UpdateCliTest.php
```diff
@@ -0,0 +1,79 @@
...
+	$wpCliMock
+		->shouldReceive('runcommand')
+		->andReturnArg(0);
+
+	$wpCliMock
+		->shouldReceive('log')
+		->andReturnArg(0);
+
+	$this->update = new UpdateCli('boilerplate');
+});
```
dingo-d, line 34: Indentation

tests/Setup/UpdateCliTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+	// ... so you can copy the example setup.json inside
+	copy(dirname(__FILE__, 3) . '/src/Setup/setup.json', dirname(__FILE__, 3) . '/cliOutput/setup.json');
+
+	// Check if an exception occured
+	$exceptionOccured = false;
+
+	$update = $this->update;
+	$update([], []);
+});
```
dingo-d, line 61: What is being tested here? I think you wanted to test something like

```php
$this->assertFalse($exceptionOccured);
```

Right? If an exception occurred, that variable would be true, and the assertion would be false.

dingo-d (changes requested):
Left one comment and some indentation issues. Great job 👍🏼

tests/Setup/UpdateCliTest.php
goranalkovic-infinum: I was trying something out and forgot to do the couple of undos to get the try/catch and the assert back. Fixed now 😄

dingo-d (approved):

iruzevic (approved):

PR #165, opened 2021-03-17
Fix the autowiring issue

dingo-d:
Exclude the non-service classes from autowiring.

This should fix the bug I was experiencing on a client's site.

This is a temp fix without tests (we'll add those in 3.1.0).

### DIFF
--- a/src/Main/Autowiring.php
+++ b/src/Main/Autowiring.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 
 namespace EightshiftLibs\Main;
 
+use EightshiftLibs\Services\ServiceInterface;
+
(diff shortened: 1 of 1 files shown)


IvanGrginovInf (approved):

volfkarlo (approved):

iobrado (approved):

IvanGrginovInf:
@dingo-d yeah I see you wrote that but we definitely need to test this on existing projects before merging to `3.1.0`

dingo-d:
@IvanGrginovInf I've added this to my project and so far didn't see any breakage 🤷🏼‍♂️

PR #167, opened 2021-03-19
updating helper

iruzevic:
the same as https://github.com/infinum/eightshift-frontend-libs/pull/306 with a few deletions that we are not using anywhere.

Tests missing. But we don't have tests for Blocks still and I will write tests for the components helpers in the next PR


src/Helpers/Components.php
```diff
@@ -378,7 +378,7 @@ public static function outputCssVariables(array $attributes, array $manifest, st
 		$name = self::camelToKebabCase($name);
 
 		foreach ($attributes as $key => $value) {
-			if (! isset($manifest['attributes'][$key]) || !isset($manifest['attributes'][$key]['variable'])) {
+			if (!  isset($manifest['attributes'][$key]) || !isset($manifest['attributes'][$key]['variable'])) {
```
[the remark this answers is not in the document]
iruzevic: Typo

src/Helpers/Components.php
```diff
@@ -422,4 +422,88 @@ public static function camelToKebabCase(string $string): string
...
+			}
+		}
+
+		// Append componentName for usage.
+		$output['componentName'] = $newNameInternal;
+
+		return $output;
+	}
+
+		/**
```
goranalkovic-infinum, line 489: Extra tab here :)

iruzevic: fixed. Not sure why the linter didn't pick this up :(

src/Helpers/Components.php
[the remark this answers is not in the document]
goranalkovic-infinum: In the left part of the condition, there's a space after the `!`, on the right condition there is no space?

src/Helpers/Components.php
goranalkovic-infinum: Happens to me as well :D

Maybe it's ignoring it because it's a multiline comment start?

src/Helpers/Components.php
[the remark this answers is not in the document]
iruzevic: Fixed

goranalkovic-infinum:
The `latest_dump.targ.gz` file got in by accident :D

goranalkovic-infinum (approved):

iruzevic:
> The `latest_dump.targ.gz` file got in by accident :D

God damit, my mind is all over the place :(. Fixed

goranalkovic-infinum:
> > The `latest_dump.targ.gz` file got in by accident :D
> 
> God damit, my mind is all over the place :(. Fixed

Let's blame it on friday :)

src/Blocks/AbstractBlocks.php
```diff
@@ -77,9 +64,10 @@ public function addThemeSupport(): void
 	 */
 	public function getBlocksDataFullRaw(): void
 	{
-		if (!$this->blocks) {
+		if (! $this->blocks) {
```
dingo-d, line 67:
```suggestion
		if (!$this->blocks) {
```

Not sure why linter isn't picking this up 🤷🏼‍♂️

src/Blocks/AbstractBlocks.php
```diff
@@ -191,12 +225,12 @@ public function render(array $attributes, string $innerBlockContent): string
 		$wrapperPath = "{$this->getWrapperPath()}/wrapper.php";
 
 		// Check if wrapper component exists.
-		if (!file_exists($wrapperPath)) {
+		if (! file_exists($wrapperPath)) {
```
dingo-d, line 228: There shouldn't be spaces after negations 🤔

src/Blocks/AbstractBlocks.php
```diff
@@ -191,12 +225,12 @@ public function render(array $attributes, string $innerBlockContent): string
...
 			throw InvalidBlock::missingWrapperViewException($wrapperPath);
 		}
 
 		// Check if actual block exists.
-		if (!file_exists($templatePath)) {
+		if (! file_exists($templatePath)) {
```
dingo-d, line 233: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -249,7 +283,7 @@ public function getCustomCategory(array $categories, \WP_Post $post): array
 	 */
 	public function renderWrapperView(string $src, array $attributes, $innerBlockContent = null): void
 	{
-		if (!file_exists($src)) {
+		if (! file_exists($src)) {
```
dingo-d, line 286: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -320,7 +354,7 @@ protected function getWrapper(): array
 	{
 		$manifestPath = "{$this->getWrapperPath()}/manifest.json";
 
-		if (!file_exists($manifestPath)) {
+		if (! file_exists($manifestPath)) {
```
dingo-d, line 357: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -343,7 +399,7 @@ protected function getComponent(string $componentName): array
 	{
 		$manifestPath = "{$this->getBlocksComponentsPath()}/{$componentName}/manifest.json";
 
-		if (!file_exists($manifestPath) && !defined('WP_CLI')) {
+		if (! file_exists($manifestPath) && !defined('WP_CLI')) {
```
dingo-d, line 402: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -365,14 +421,14 @@ protected function getSettings(): array
 	{
 		$manifestPath = "{$this->getBlocksPath()}/manifest.json";
 
-		if (!file_exists($manifestPath)) {
+		if (! file_exists($manifestPath)) {
```
dingo-d, line 424: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -365,14 +421,14 @@ protected function getSettings(): array
...
 			throw InvalidBlock::missingSettingsManifestException($manifestPath);
 		}
 
 		$settings = implode(' ', (array)file(($manifestPath)));
 		$settings = json_decode($settings, true);
 
-		if (!isset($settings['namespace'])) {
+		if (! isset($settings['namespace'])) {
```
dingo-d, line 431: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -457,7 +513,7 @@ protected function prepareComponentAttributes(array $blockDetails): array
 		$output = [];
 		$componentAttributes = [];
 
-		if (!isset($blockDetails['components'])) {
+		if (! isset($blockDetails['components'])) {
```
dingo-d, line 516: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -486,6 +542,95 @@ protected function prepareComponentAttributes(array $blockDetails): array
...
+		foreach ($componentsList as $key => $value) {
+			$items = array_values(
+				array_filter(
+					$components,
+					function ($item) use ($value) {
+						$componentName = $item['componentName'] ?? '';
+
+						if ($componentName === $value) {
+							return $item;
+						}
```
dingo-d, line 617: This is not the correct usage of `array_filter`. The `array_filter` will go through every item in the array, check the condition, and if the condition is `true` it will keep the item in the array. If not, it will remove it.

So if your intention is to just remove the items, then the only thing you need here is:

```suggestion
						return $componentName === $value;
```

If you need to return a different thing in the array, then you need `array_map`.

src/Blocks/AbstractBlocks.php
```diff
@@ -501,19 +646,19 @@ function (string $blockPath) {
 
 				$block = $this->parseManifest($block);
 
-				if (!isset($block['blockName'])) {
+				if (! isset($block['blockName'])) {
```
dingo-d, line 649: Unnecessary space

src/Blocks/AbstractBlocks.php
```diff
@@ -501,19 +646,19 @@ function (string $blockPath) {
...
 					throw InvalidBlock::missingNameException($blockPath);
 				}
 
-				if (!isset($block['classes'])) {
+				if (! isset($block['classes'])) {
```
dingo-d, line 653: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -501,19 +646,19 @@ function (string $blockPath) {
...
 					$block['classes'] = [];
 				}
 
-				if (!isset($block['attributes'])) {
+				if (! isset($block['attributes'])) {
```
dingo-d, line 657: ☝🏼

src/Blocks/AbstractBlocks.php
```diff
@@ -501,19 +646,19 @@ function (string $blockPath) {
...
 					$block['attributes'] = [];
 				}
 
-				if (!isset($block['hasInnerBlocks'])) {
+				if (! isset($block['hasInnerBlocks'])) {
```
dingo-d, line 661: ☝🏼

src/Blocks/BlocksExample.php
```diff
@@ -64,6 +75,21 @@ protected function getBlocksPath(): string
...
+	 * @param array  $attributes Object of attributes from block/component.
+	 * @param string $realName Old key to use, generally this is the name of the block/component.
+	 * @param string $newName New key to use to rename attributes.
+	 * @param bool   $isBlock Check if helper is used on block or component.
+	 *
+	 * @return array
+	 */
+	public static function props(array $attributes, string $realName, string $newName = '', bool $isBlock = false): array
+	{
+		return Components::props($attributes, $realName, $newName, $isBlock, \apply_filters(self::BLOCKS_DEPENDENCY_FILTER_NAME, 'dependency'));
```
dingo-d, line 90: Above this is `static`, here it's `self`. This is intentional? As in the filter name should remain what's defined here, so that it can be referenced to and not changed in the child class?

src/Helpers/Components.php
```diff
@@ -112,7 +112,7 @@ public static function render(string $component, array $attributes = [], string
 			}
 		}
 
-		if (!file_exists($componentPath)) {
+		if (! file_exists($componentPath)) {
```
dingo-d, line 115: Space

src/Helpers/Components.php
```diff
@@ -173,7 +173,7 @@ public static function getManifest(string $path): array
 
 		$manifest = "{$path}/manifest.json";
 
-		if (!file_exists($manifest)) {
+		if (! file_exists($manifest)) {
```
dingo-d, line 176: ☝🏼

src/Helpers/Components.php
```diff
@@ -277,11 +277,11 @@ public static function selector($condition, string $block, string $element = '',
 		$modifier = trim($modifier);
 		$block = trim($block);
 
-		if (!empty($element)) {
+		if (! empty($element)) {
```
dingo-d, line 280: ☝🏼

src/Helpers/Components.php
```diff
@@ -277,11 +277,11 @@ public static function selector($condition, string $block, string $element = '',
...
 			$fullElement = "__{$element}";
 		}
 
-		if (!empty($modifier)) {
+		if (! empty($modifier)) {
```
dingo-d, line 284: ☝🏼

src/Helpers/Components.php
```diff
@@ -298,7 +298,7 @@ public static function outputCssVariablesGlobal(array $globalManifest): string
 	{
 		$output = '';
 
-		if (!$globalManifest || !isset($globalManifest['globalVariables'])) {
+		if (! $globalManifest || ! isset($globalManifest['globalVariables'])) {
```
dingo-d, line 301: ☝🏼

src/Helpers/Components.php
```diff
@@ -369,7 +369,7 @@ public static function outputCssVariables(array $attributes, array $manifest, st
 	{
 		$output = '';
 
-		if (!$attributes || !$manifest) {
+		if (! $attributes || ! $manifest) {
```
dingo-d, line 372: ☝🏼

src/Helpers/Components.php
```diff
@@ -378,7 +378,7 @@ public static function outputCssVariables(array $attributes, array $manifest, st
 		$name = self::camelToKebabCase($name);
 
 		foreach ($attributes as $key => $value) {
-			if (! isset($manifest['attributes'][$key]) || !isset($manifest['attributes'][$key]['variable'])) {
+			if (! isset($manifest['attributes'][$key]) || ! isset($manifest['attributes'][$key]['variable'])) {
```
dingo-d, line 381: ☝🏼

src/Helpers/Components.php
```diff
@@ -422,4 +422,88 @@ public static function camelToKebabCase(string $string): string
...
+	 *
+	 * @return array
+	 */
+	public static function props(array $attributes, string $realName, string $newName = '', bool $isBlock = false, array $globalData): array // phpcs:ignore PEAR.Functions.ValidDefaultValue.NotAtEnd
+	{
+
+		$newNameInternal = $newName;
+
+		// Check if newName key is passed if not use the default one from block/component name.
+		if (! $newName) {
```
dingo-d, line 443:
```suggestion
		if (!$newName) {
```

src/Helpers/Components.php
```diff
@@ -422,4 +422,88 @@ public static function camelToKebabCase(string $string): string
...
+		// If component use components dependency tree.
+		$dependency = $globalData['components'][$realName];
+
+		// If block use blocks dependency tree.
+		if ($isBlock) {
+			$dependency = $globalData['blocks'][$realName];
+		}
+
+		// If dependency is empty put the name in the array for the easier checks later on.
+		if (! $dependency) {
```
dingo-d, line 458:
```suggestion
		if (!$dependency) {
```

src/Helpers/Components.php
```diff
@@ -422,4 +422,88 @@ public static function camelToKebabCase(string $string): string
...
+			$dependency = [$newNameInternal];
+		}
+
+		foreach ($attributes as $key => $value) {
+			$result = false;
+			foreach ($dependency as $element) {
+				if ($element === substr($key, 0, strlen($element))) {
+					$result =  true;
```
dingo-d, line 466: This `$result` will get rewritten for every loop. Not sure what was the intention here 🤔 

You'd have to move the below conditional inside the loop for this to make sense, and in that case the entire conditional body could be put instead of the `$result`.

dingo-d (changes requested):
Left some comments. Not sure why the coding standards aren't picking up space in front of the `!`. There should be no space before the negation operator.

src/Blocks/BlocksExample.php
iruzevic: Yes, this should not be changed in the child class.

src/Helpers/Components.php
iruzevic: Basically this serves as a test that should go in the if statement below but since it ugly in PHP to put this kind of code in condition I have written  this like so. Basically this part ($result) can be extracted in an external helper function.

This represents this in the JS
https://github.com/infinum/eightshift-frontend-libs/blob/a7ea385994c8370dd8948cd48fec337c80b0d79b/scripts/editor/props.js#L49

Do you have any idea how to write this better? The current code works as expected

src/Blocks/AbstractBlocks.php
iruzevic: Fixed

iruzevic:
Fixed all the spaces after ! in all the files in the project

src/Helpers/Components.php
dingo-d: Oh, I just noticed that there are 2 foreach loops! I mean you could try to use `array_search(substr($key, 0, strlen($element)), $dependency)` and use that as a `$result`. It's a bit hard to visualize this without example data (I have no idea what `$attributes` look like here).

iruzevic: Is it ok with you to leave it like this for now until I write tests and have everything working so we can make changes safely?

dingo-d (approved):

PR #168, opened 2021-03-24
Feature/exception tests

iobrado:
Still very much WIP, I've added tests for following classes:
- ComponentException
- InvalidBlock
- InvalidCallback

Also, PR contains few docblock description updates.
In case the approach for the tests is wrong, please provide me with an example how it should be done. Since most tests are quite similar, it shouldn't take too long to fix them.

### DIFF
--- a/tests/Exception/ComponentExceptionTest.php
+++ b/tests/Exception/ComponentExceptionTest.php
@@ -5,46 +5,39 @@ namespace Tests\Unit\Exception;
 use EightshiftLibs\Exception\ComponentException;
 use stdClass;
 
+use Brain\Monkey;
+
 use function Tests\setupMocks;
 
 beforeAll(function () {
+	Monkey\setUp();
 	setupMocks();
 });
 
-test('Throws exception if ensure_string argument is invalid.', function () {
-
-	$object = new stdClass();
-	$integer = 7;
-	$null = null;
-	$bool = true;
-
-	$exceptionObject = ComponentException::throwNotStringOrArray($object);
-	$exceptionInteger = ComponentException::throwNotStringOrArray($integer);
-	$exceptionNull = ComponentException::throwNotStringOrArray($null);
-	$exceptionBool = ComponentException::throwNotStringOrArray($bool);
-
-	$this->assertIsObject($exceptionObject);
-	$this->assertIsObject($exceptionInteger);
-	$this->assertIsObject($exceptionNull);
-	$this->assertIsObject($exceptionBool);
-	
-	$this->assertObjectHasAttribute('message', $exceptionObject);
-	$this->assertObjectHasAttribute('message', $exceptionInteger);
-	$this->assertObjectHasAttribute('message', $exceptionNull);
-	$this->assertObjectHasAttribute('message', $exceptionBool);
-
-	$this->assertStringContainsString('Object couldn&#039;t be converted to string. Please provide only string or array.', $exceptionObject->getMessage());
-	$this->assertStringContainsString("{$integer} variable is not a string or array but rather integer", $exceptionInteger->getMessage());
-	$this->assertStringContainsString("{$null} variable is not a string or array but rather NULL", $exceptionNull->getMessage());
-	$this->assertStringContainsString("{$bool} variable is not a string or array but rather boolean", $exceptionBool->getMessage());
+afterAll(function() {
+	Monkey\tearDown();
 });
 
-test('Throws exception if unable to locate component.', function () {
+test('Checks if the throwNotStringOrArray method functions correctly.',
+	function ($argument) {
+		$exception = ComponentException::throwNotStringOrArray($argument);
+		$type = gettype($argument);
+
+		$this->assertIsObject($exception, "The {$exception} should be an instance of ComponentException class");
+		$this->assertObjectHasAttribute('message', $exception, "Object doesn't contain message attribute");
+		// $this->assertStringContainsString("{$argument} variable is not a string or array but rather {$type}", $exception->getMessage(), "Strings for message if item is {$type} do not match!");
+
+	})
+	->with('exceptionArguments');
+
+	// $this->assertStringContainsString('Object couldn&#039;t be converted to string. Please provide only string or array.', $exceptionObject->getMessage(), "Strings for 'Object couldn't be converted to string' message do not match!");
+
+test('Checks if throwUnableToLocateComponent method will return correct response.', function () {
 
 	$component = 'nonexistent';
 	$output = ComponentException::throwUnableToLocateComponent($component);
 
-	$this->assertIsObject($output);
-	$this->assertObjectHasAttribute('message', $output);
-	$this->assertStringContainsString("Unable to locate component by path: {$component}", $output->getMessage());
+	$this->assertIsObject($output, "The {$output} should be an instance of ComponentException class");
+	$this->assertObjectHasAttribute('message', $output, "Object doesn't contain message attribute");
+	$this->assertStringContainsString("Unable to locate component by path: {$component}", $output->getMessage(), "Strings for 'Unable to locate component by path' message do not match!");
 });
--- a/tests/Exception/ComponentExceptionTest.php
+++ b/tests/Exception/ComponentExceptionTest.php
@@ -20,17 +20,27 @@ afterAll(function() {
 
 test('Checks if the throwNotStringOrArray method functions correctly.',
 	function ($argument) {
-		$exception = ComponentException::throwNotStringOrArray($argument);
+		$exceptionObject = ComponentException::throwNotStringOrArray($argument);
 		$type = gettype($argument);
 
-		$this->assertIsObject($exception, "The {$exception} should be an instance of ComponentException class");
-		$this->assertObjectHasAttribute('message', $exception, "Object doesn't contain message attribute");
-		// $this->assertStringContainsString("{$argument} variable is not a string or array but rather {$type}", $exception->getMessage(), "Strings for message if item is {$type} do not match!");
+		$this->assertIsObject($exceptionObject, "The {$exceptionObject} should be an instance of ComponentException class");
+		$this->assertObjectHasAttribute('message', $exceptionObject, "Object doesn't contain message attribute");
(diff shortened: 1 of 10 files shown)


tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
dingo-d, line 1: You didn't have to test this one, since it's also covered, but I guess it doesn't hurt to have exceptions covered on their own 😄 

Screenshot 2021-03-24 at 10 32 00

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+	
+	$this->assertObjectHasAttribute('message', $exceptionObject);
+	$this->assertObjectHasAttribute('message', $exceptionInteger);
+	$this->assertObjectHasAttribute('message', $exceptionNull);
+	$this->assertObjectHasAttribute('message', $exceptionBool);
+
+	$this->assertStringContainsString('Object couldn&#039;t be converted to string. Please provide only string or array.', $exceptionObject->getMessage());
+	$this->assertStringContainsString("{$integer} variable is not a string or array but rather integer", $exceptionInteger->getMessage());
+	$this->assertStringContainsString("{$null} variable is not a string or array but rather NULL", $exceptionNull->getMessage());
+	$this->assertStringContainsString("{$bool} variable is not a string or array but rather boolean", $exceptionBool->getMessage());
```
dingo-d, line 39: When having multiple same assertions in a test, it's a good practice to add the test message at the end, so that, in case something fails, you'll know immediately what failed and why.

For instance, in the first block I'd add something like

```php
	$this->assertIsObject($exceptionObject, "The {$exceptionObject} should be an instance of ComponentException class");
	$this->assertIsObject($exceptionInteger, "The {$exceptionInteger} should be an instance of ComponentException class");
	$this->assertIsObject($exceptionNull, "The {$exceptionNull} should be an instance of ComponentException class");
	$this->assertIsObject($exceptionBool, "The {$exceptionBool} should be an instance of ComponentException class");
```

That way it's easier to identify what failed and where.

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+use EightshiftLibs\Exception\ComponentException;
+use stdClass;
+
+use function Tests\setupMocks;
+
+beforeAll(function () {
+	setupMocks();
+});
+
+test('Throws exception if ensure_string argument is invalid.', function () {
```
dingo-d, line 14: The name of the test is a bit misleading, as you are not acutally testing the `ensure_string` helper, but rather the `throwNotStringOrArray` method of the exception.

So I'd rename this to something like `Checks if the throwNotStringOrArray method functions correctly.`

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+	$this->assertIsObject($exceptionInteger);
+	$this->assertIsObject($exceptionNull);
+	$this->assertIsObject($exceptionBool);
+	
+	$this->assertObjectHasAttribute('message', $exceptionObject);
+	$this->assertObjectHasAttribute('message', $exceptionInteger);
+	$this->assertObjectHasAttribute('message', $exceptionNull);
+	$this->assertObjectHasAttribute('message', $exceptionBool);
+
+	$this->assertStringContainsString('Object couldn&#039;t be converted to string. Please provide only string or array.', $exceptionObject->getMessage());
```
dingo-d, line 36:
```suggestion
	$this->assertStringContainsString('Object couldn\'t be converted to string. Please provide only string or array.', $exceptionObject->getMessage());
```

Just escape the `'` character if you are using it inside a string 😉

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+	$this->assertStringContainsString("{$integer} variable is not a string or array but rather integer", $exceptionInteger->getMessage());
+	$this->assertStringContainsString("{$null} variable is not a string or array but rather NULL", $exceptionNull->getMessage());
+	$this->assertStringContainsString("{$bool} variable is not a string or array but rather boolean", $exceptionBool->getMessage());
+});
+
+test('Throws exception if unable to locate component.', function () {
```
dingo-d, line 42: Same as the above, the description should be a bit different. Since you are just testing the exceptions themselves.

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+beforeAll(function () {
+	setupMocks();
+});
+
+test('Throws exception if ensure_string argument is invalid.', function () {
+
+	$object = new stdClass();
+	$integer = 7;
+	$null = null;
+	$bool = true;
```
dingo-d, line 19: These could be set inside datasets. Check the [file](https://github.com/infinum/eightshift-libs/blob/release/3.1.0/tests/Datasets/Arguments.php), and [example](https://github.com/infinum/eightshift-libs/blob/release/3.1.0/tests/Helpers/ComponentHelpersTest.php#L39-L44) of their usage.

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
...
+});
+
+test('Throws exception if unable to locate component.', function () {
+
+	$component = 'nonexistent';
+	$output = ComponentException::throwUnableToLocateComponent($component);
+
+	$this->assertIsObject($output);
+	$this->assertObjectHasAttribute('message', $output);
+	$this->assertStringContainsString("Unable to locate component by path: {$component}", $output->getMessage());
```
dingo-d, line 49: Same comment about adding extra descriptions. Your future self will be glad you added them 😄

tests/Exception/ComponentExceptionTest.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Tests\Unit\Exception;
+
+use EightshiftLibs\Exception\ComponentException;
+use stdClass;
+
+use function Tests\setupMocks;
+
+beforeAll(function () {
```
dingo-d, line 10: Make sure you add `Monkey\setUp();` before calling the `setupMocks()` and then add

```php
afterAll(function() {
	Monkey\tearDown();
});
```

tests/Exception/InvalidBlockTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Tests\Unit\Exception;
+
+use EightshiftLibs\Exception\InvalidBlock;
+
+use function Tests\setupMocks;
+
+beforeAll(function () {
```
dingo-d, line 9: Same comment about using `beforeAll` and `afterAll` in the test

tests/Exception/InvalidBlockTest.php
```diff
@@ -0,0 +1,109 @@
...
+	setupMocks();
+});
+
+test('Throws error if blocks are missing.', function () {
+
+	$missingBlocks = InvalidBlock::missingBlocksException();
+
+	$this->assertIsObject($missingBlocks);
+	$this->assertObjectHasAttribute('message', $missingBlocks);
+	$this->assertStringContainsString('There are no blocks added in your project.', $missingBlocks->getMessage());
```
dingo-d, line 19: Multiple asserts without comments 🙂

tests/Exception/InvalidBlockTest.php
```diff
@@ -0,0 +1,109 @@
...
+});
+
+test('Throws error if manifest key blockName is missing.', function () {
+
+	$blockPath = 'some/random/path';
+	$missingName = InvalidBlock::missingNameException($blockPath);
+
+	$this->assertIsObject($missingName);
+	$this->assertObjectHasAttribute('message', $missingName);
+	$this->assertStringContainsString("Block in this path {$blockPath} is missing blockName key in its manifest.json.", $missingName->getMessage());
```
dingo-d, line 38: ☝🏼

tests/Exception/InvalidBlockTest.php
```diff
@@ -0,0 +1,109 @@
...
+	$this->assertStringContainsString("Wrapper view is missing. Template should be located in this path {$wrapperPath}", $missingWrapperView->getMessage());
+});
+
+test('Throws error if global manifest settings key namespace is missing.', function () {
+
+	$missingNamespace = InvalidBlock::missingNamespaceException();
+
+	$this->assertIsObject($missingNamespace);
+	$this->assertObjectHasAttribute('message', $missingNamespace);
+	$this->assertStringContainsString('Global Blocks settings manifest.json is missing a key called namespace. This key prefixes all block names.', $missingNamespace->getMessage());
```
dingo-d, line 108: Just add the messages for every test here 😄

tests/Exception/InvalidCallbackTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+use EightshiftLibs\Exception\InvalidCallback;
+
+use function Tests\setupMocks;
+
+beforeAll(function () {
+	setupMocks();
+});
+
+test('Create a new instance of the exception for a callback class name that is not recognized.', function () {
```
dingo-d, line 13: Again, the name is a bit odd. You are testing if the `fromCallback` method will return correct response.

dingo-d (changes requested):
Great job 👍🏼 

I've left some comments on how you can improve the tests a bit.

iobrado:
Thanks for very detailed and insightful comments 🙂  I'll update the tests according to your suggestions.

iobrado:
I've added all the other tests and did fixes as you suggested. The only thing I'm still having trouble is in CompoentExceptionTest.php on line 41. I've left that one commented out so maybe you can figure out what's causing that test to fail.

dingo-d (approved):

dingo-d:
@iobrado I'm dumb. Can you uncomment that test and add the `&#039;` instead of the `'`. The `esc_html` will escape the special character. 🙈

iobrado:
> @iobrado I'm dumb. Can you uncomment that test and add the `&#039;` instead of the `'`. The `esc_html` will escape the special character. 🙈

Sure thing, I've updated the code accordingly and pushed the changes 🙂  if everything else is fine, we can merge this.

iruzevic (approved):

PR #172, opened 2021-03-24
Blocks Tests

iruzevic:
Adding tests for the block


goranalkovic-infinum (approved):

tests/Blocks/BlocksExampleTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+afterEach(function() {
+	Monkey\tearDown();
+});
+
+test('Register method will call init hooks', function () {
+	$this->blocks->register();
+
+	$this->assertSame(10, has_action('init', 'EightshiftBoilerplate\Blocks\BlocksExample->getBlocksDataFullRaw()'));
+	$this->assertSame(11, has_action('init', 'EightshiftBoilerplate\Blocks\BlocksExample->registerBlocks()'));
```
dingo-d, line 26: Add an error message in case a test fails so that you can easily differentiate which test failed. For example:

```php
$this->assertSame(10, has_action('init', 'EightshiftBoilerplate\Blocks\BlocksExample->getBlocksDataFullRaw()'), 'The callback getBlocksDataFullRaw should be hooked to init hook with priority 10');
```

Similar should be done with the second one.

Every time you have multiple of the same assertion types (`assertTrue`, `assertStringContainsString`, etc.) it's good to have error message so in case something fails you can find it immediately.

tests/Blocks/BlocksExampleTest.php
```diff
@@ -0,0 +1,52 @@
...
+	$this->blocks->register();
+
+	$this->assertSame(10, has_filter('block_categories', 'EightshiftBoilerplate\Blocks\BlocksExample->getCustomCategory()'));
+});
+
+test('Register method will call after_setup_theme hooks', function () {
+	$this->blocks->register();
+
+	$this->assertSame(25, has_action('after_setup_theme', 'EightshiftBoilerplate\Blocks\BlocksExample->addThemeSupport()'));
+	$this->assertSame(11, has_action('after_setup_theme', 'EightshiftBoilerplate\Blocks\BlocksExample->changeEditorColorPalette()'));
```
dingo-d, line 39: ☝🏼

dingo-d (changes requested):
Left one comment

tests/Blocks/BlocksExampleTest.php
iruzevic: Is it better to maybe a add different tests for each hooks to avoid this issue?

dingo-d: Nah, since you are just testing hooks functionality in the same class. It's always recommended to add the description at the end when using multiple same assertions inside one test.

iruzevic: fixed

tests/Blocks/BlocksExampleTest.php
iruzevic: fixed

tests/Blocks/BlocksExampleTest.php
```diff
@@ -12,41 +13,136 @@
...
+
+	$post = $this->wpFaker->post();
+
+	$blocks = $this->blocksExample->getAllBlocksList(true, $post);
+
+	$this->assertEquals(true, $blocks);
+
+	$blocks = $this->blocksExample->getAllBlocksList(false, $post);
+
+	$this->assertEquals(false, $blocks);
```
dingo-d, line 88: Can you add an error message, in case this fails

tests/Blocks/BlocksExampleTest.php
```diff
@@ -12,41 +13,136 @@
...
+test('Asserts that getAllBlocksList will return only projects blocks.', function () {
+
+	$post = $this->wpFaker->post();
+	$list = $this->blocksExample->getAllBlocksList([], $post);
+
+	$this->assertIsArray($list);
+	$this->assertNotContains('core/paragraph', $list);
+	$this->assertContains('eightshift-boilerplate/button', $list);
+	$this->assertContains('core/block', $list);
+	$this->assertContains('core/template', $list);
```
dingo-d, line 100: Here as well

tests/Blocks/BlocksExampleTest.php
```diff
@@ -12,41 +13,136 @@
...
+
+	$items = $this->blocksExample->getBlocksDataFullRawItem('');
+
+	$this->assertIsArray($items);
+	$this->assertNotContains('button', array_keys($items));
+	$this->assertContains('dependency', array_keys($items));
+	$this->assertContains('blocks', array_keys($items));
+	$this->assertContains('components', array_keys($items));
+	$this->assertContains('wrapper', array_keys($items));
+	$this->assertContains('settings', array_keys($items));
```
dingo-d, line 113: ☝🏼

tests/Blocks/BlocksExampleTest.php
```diff
@@ -12,41 +13,136 @@
...
+
+test('Asserts that getBlocksDataFullRawItem will return empty array if code is run using WP_CLI.', function () {
+
+	define('WP_CLI', true);
+
+	$items = $this->blocksExample->getBlocksDataFullRawItem();
+
+	$this->assertIsArray($items);
+	$this->assertEmpty($items);
+	$this->assertEmpty($items);
```
dingo-d, line 146: ☝🏼

dingo-d (changes requested):
Left a few comments. The wpfaker is cool, although faker lib has been archived so this is something we'll need to keep in mind when going forward (https://marmelab.com/blog/2020/10/21/sunsetting-faker.html)

iruzevic:
> Left a few comments. The wpfaker is cool, although faker lib has been archived so this is something we'll need to keep in mind when going forward (https://marmelab.com/blog/2020/10/21/sunsetting-faker.html)

@dingo-d 
Agreed, if there is some other way to mimic WP_Post object let me know and I will remove this dependency.
This i used here and it is not used in the actual implementation. 

Screenshot 2021-04-01 at 09 45 27

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -88,6 +97,8 @@ function ($argument) {
 })->throws(ComponentException::class);
 
 test('Asserts that render used components defaults', function () {
+	Functions\when('get_template_directory')->justReturn(dirname(__FILE__, 2) . '/data/frontend-libs');
```
gabriel-glo, line 100: This mock could be extracted to the setup function to avoid repeating.

IvanGrginovInf:
> > Left a few comments. The wpfaker is cool, although faker lib has been archived so this is something we'll need to keep in mind when going forward (https://marmelab.com/blog/2020/10/21/sunsetting-faker.html)
> 
> @dingo-d
> Agreed, if there is some other way to mimic WP_Post object let me know and I will remove this dependency.
> This i used here and it is not used in the actual implementation.
> 
> Screenshot 2021-04-01 at 09 45 27

@iruzevic I mean you can mock it the same way you mock any other class using `Mockery`. Example can be found in rest routes https://github.com/infinum/eightshift-libs/blob/release/3.1.0/tests/Rest/Routes/RouteExampleTest.php

1. Mock native WP object: `$this->wpRestRequest = \Mockery::mock('alias:WP_REST_Request');`
2. Pass it as argument to any method: $output = `$this->route->routeCallback($this->wpRestRequest);`
3. (Optionally) if you need to mock any WP_Post properties / methods: 

```php 
$this->wpRestRequest
		->shouldReceive('get_body')
		->andReturn(json_encode([$this->mockRequestKey => $this->mockRequestValue]));
```

tests/Helpers/ComponentHelpersTest.php
dingo-d: There already is an implementation `Functions\when('get_template_directory')->justReturn(dirname(__FILE__) . '/data');` inside the `setupMocks()` function.

iruzevic:
> 1. $this->wpPost = \Mockery::mock('WP_Post');

awesome tnx this worked

dingo-d (approved):
👍🏼  Great job!

gabriel-glo (approved):

goranalkovic-infinum (approved):

PR #173, opened 2021-03-25
Autowiring fixes + tests

IvanGrginovInf:
Changelog
---
* Added tests for `Autowiring.php`
* Added a bunch of Mock classes to simulate all variations that Autowiring supports
* Throw a helpful exception if it finds a file that looks like it should be PSR-4 compliant but the namespace is wrong (it ignores blocks, scripts, and other things)
* Throw exception if you use interface-based dependency and provide a filename from which we're unable to guess which class to inject
* Throw exception if you use interface-based dependency and we find more than 1 class with the same name (based on the variable you're using) that implement that same interface
* Throw exception if you use primitive values as dependencies in Services (or dependencies of those services)
* All manually defined dependency trees are ignored in the Autowiring process (so you can fix any of the above exceptions by also manually defining that service dependency tree - like the old way)

Please see the descriptions of tests to see what is covered. If you wish to test your use-case, you can create your classes/interfaces inside `tests/Datasets/Autowiring` and run the tests.

### DIFF
--- a/src/Exception/InvalidAutowireDependency.php
+++ b/src/Exception/InvalidAutowireDependency.php
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * File containing the failure exception class when trying to inject interface based dependencies into a class (using autowiring) but
+ * failing to provide the correct variable name by which we can find a class to inject.
+ *
+ * @package EightshiftLibs\Exception
+ */
+
+declare(strict_types=1);
+
+namespace EightshiftLibs\Exception;
+
+/**
+ * Class Component_Exception.
+ */
+final class InvalidAutowireDependency extends \InvalidArgumentException implements GeneralExceptionInterface
+{
+
+	/**
+	 * Throws exception if we cant guess the class to inject.
+	 *
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwUnableToFindClass(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Unable to find \"%s\" class that implements %s (looking in \$filenameIndex). When injecting Interface dependencies, please make sure your variable name in __construct() matches the filename of a class implementing that interface (otherwise we dont know which class to inject). Alternatively you can define the dependency tree manually for this class using \$main->getServiceClasses(). See https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
+				$className,
+				$interfaceName
+			)
+		);
+	}
+
+	/**
+	 * Throws exception if we cant guess the class to inject because we found more than 1 with same name that implement $interfaceName.
+	 *
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwMoreThanOneClassFound(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Found more than 1 class called \"%s\" that implements %s interface. Please make sure you dont have more than 1 class with the same name implementing the same interface. Alternatively you can manually defined dependencies for the class that uses %s interface as a dependency. See: https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
+				$className,
+				$interfaceName,
+				$interfaceName
+			)
+		);
+	}
+
+	/**
+	 * Throws exception if we find a primitive dependency on a class that's not been manually built.
+	 *
+	 * @param string $className Class name we're looking for.
+	 * @return static
+	 */
+	public static function throwPrimitiveDependencyFound(string $className)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Found a primitive dependency for %s. Autowire is unable to figure out what value needs to be injected here. Please define the dependency tree for this class manually using \$main->getServiceClasses(). See: https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-a-primitive-parameter-string-int-etc-inside-a-constructor-method",
+				$className
+			)
+		);
+	}
+}
--- a/src/Main/Autowiring.php
+++ b/src/Main/Autowiring.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 
 namespace EightshiftLibs\Main;
 
+use EightshiftLibs\Services\ServiceInterface;
+
 /**
  * The file that defines the autowiring process
  */
@@ -56,8 +58,13 @@ class Autowiring
 				continue;
 			}
 
-			// Skip abstract classes, interfaces & traits.
-			if ($reflClass->isAbstract() || $reflClass->isInterface() || $reflClass->isTrait()) {
+			// Skip abstract classes, interfaces & traits, and non service classes.
+			if (
+				$reflClass->isAbstract() ||
+				$reflClass->isInterface() ||
+				$reflClass->isTrait() ||
+				!$reflClass->implementsInterface(ServiceInterface::class)
+			) {
 				continue;
 			}
 
--- a/src/Main/Autowiring.php
+++ b/src/Main/Autowiring.php
@@ -105,14 +105,16 @@ class Autowiring
 	 * @param array  $filenameIndex Filename index. Maps filenames to class names.
 	 * @param array  $classInterfaceIndex Class interface index. Maps classes to interfaces they implement.
 	 *
-	 * @throws \ReflectionException Exception thrown in case class is missing.
-	 * @throws \Exception If things we're looking for are missing
-	 *                    inside filename or classInterface index (which shouldn't happen).
+	 * @throws InvalidAutowireDependency If a primitive dependency is found.
 	 *
 	 * @return array
 	 */
 	private function buildDependencyTree(string $relevantClass, array $filenameIndex, array $classInterfaceIndex): array
 	{
+		// Keeping PHPStan happy.
+		if (!class_exists($relevantClass, false)) {
+			return [];
+		}
 
 		$dependencyTree = [];
 		$reflClass = new \ReflectionClass($relevantClass);
@@ -122,6 +124,8 @@ class Autowiring
 			// Go through each constructor parameter.
 			foreach ($reflClass->getConstructor()->getParameters() as $reflParam) {
 				$type = $reflParam->getType();
+
+				// Skip parameters without type hints.
 				if ($type instanceof \ReflectionNamedType) {
 					$className = $type->getName();
 					$isBuiltin = $type->isBuiltin();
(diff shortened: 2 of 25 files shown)


src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+namespace EightshiftLibs\Exception;
+
+/**
+ * Class Component_Exception.
+ */
+final class InvalidAutowireDependency extends \InvalidArgumentException implements GeneralExceptionInterface
+{
+
+	/**
+	 * Throws exception if we cant guess the class to inject.
```
dingo-d, line 21:
```suggestion
	 * Throws an exception if we can't guess the class to inject.
```

src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Unable to find \"%s\" class that implements %s (looking in \$filenameIndex). When injecting Interface dependencies, please make sure your variable name in __construct() matches the filename of a class implementing that interface (otherwise we dont know which class to inject). Alternatively you can define the dependency tree manually for this class using \$main->getServiceClasses(). See https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
+				$className,
+				$interfaceName
+			)
+		);
+	}
+
+	/**
+	 * Throws exception if we cant guess the class to inject because we found more than 1 with same name that implement $interfaceName.
```
dingo-d, line 40:
```suggestion
	 * Throws an exception if we can't guess the class to inject because we found more than 1 with the same name that implements $interfaceName.
```

src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwMoreThanOneClassFound(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Found more than 1 class called \"%s\" that implements %s interface. Please make sure you dont have more than 1 class with the same name implementing the same interface. Alternatively you can manually defined dependencies for the class that uses %s interface as a dependency. See: https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
```
dingo-d, line 51:
```suggestion
				"Found more than 1 class called \"%s\" that implements %s interface. Please make sure you don't have more than 1 class with the same name implementing the same interface. Alternatively, you can manually define dependencies for the class that uses the %s interface as a dependency. See: https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
```

src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwUnableToFindClass(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
+				"Unable to find \"%s\" class that implements %s (looking in \$filenameIndex). When injecting Interface dependencies, please make sure your variable name in __construct() matches the filename of a class implementing that interface (otherwise we dont know which class to inject). Alternatively you can define the dependency tree manually for this class using \$main->getServiceClasses(). See https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
```
dingo-d, line 32:
```suggestion
				"Unable to find \"%1$s\" class that implements %2$s (looking in \$filenameIndex). When injecting Interface dependencies, please make sure your variable name in __construct() matches the filename of a class implementing that interface (otherwise we don't know which class to inject). Alternatively you can define the dependency tree manually for this class using \$main->getServiceClasses(). See https://infinum.github.io/eightshift-docs/docs/basics/autowiring#what-if-my-class-has-an-interface-parameter-inside-the-constructor-method",
```

src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+	 *
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwUnableToFindClass(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
```
dingo-d, line 31:
```suggestion
				/* translators: 1: The class name, 2: The interface name. */
```

src/Exception/InvalidAutowireDependency.php
```diff
@@ -0,0 +1,75 @@
...
+	 *
+	 * @param string $className Class name we're looking for.
+	 * @param string $interfaceName Class we're looking for needs to implement this.
+	 * @return static
+	 */
+	public static function throwMoreThanOneClassFound(string $className, string $interfaceName)
+	{
+		return new static(
+			sprintf(
+				/* translators: %s is replaced with the className and interfaceName. */
```
dingo-d, line 50:
```suggestion
				/* translators: 1: The class name, 2: The interface name. */
```

src/Main/Autowiring.php
```diff
@@ -89,54 +105,64 @@ public function buildServiceClasses(): array
...
-					// If we're unable to find exactly 1 class for whatever reason, just skip it, the user
-					// will have to define the dependencies manually.
-					if (empty($matchedClass)) {
-						continue;
+				// Keeping PHPStan happy.
+				if (class_exists($className, false) || interface_exists($className, false)) {
+					$reflClassForParam = new \ReflectionClass($className);
+
+					// If the expected type is interface, try guessing based on var name. Otherwise
+					// Just inject that class.
```
dingo-d, line 148:
```suggestion
					// If the expected type is an interface, try guessing based on var name.
					// Otherwise just inject that class.
```

src/Main/Autowiring.php
```diff
@@ -218,10 +246,11 @@ private function tryToFindMatchingClass(
 		array $filenameIndex,
 		array $classInterfaceIndex
 	): string {
-		// If there's no matches in filename index by variable, we need to skip it, this dependency's definition.
-		// list need sto be build manually.
+		// If there's no matches in filename index by variable, we need to throw an exception to let the user
+		// know he either needs to provide the correct variable name OR manually define the dependencies for this class.
```
dingo-d, line 250:
```suggestion
		// know they either needs to provide the correct variable name OR manually define the dependencies for this class.
```

src/Main/Autowiring.php
```diff
@@ -275,25 +307,24 @@ private function buildFilenameIndex(array $allRelevantClasses): array
...
 		$classInterfaceIndex = [];
-		foreach ($allRelevantClasses as $relevantClass) {
+		foreach ($reflectionClasses as $projectClass => $reflectionClass) {
 			$interfaces = array_map(
 				function () {
 					return true;
 				},
-				(new \ReflectionClass($relevantClass))->getInterfaces()
+				$reflectionClass->getInterfaces()
 			);
```
dingo-d, line 325: This just goes through the reflection classes and tries to check if the classes have interfaces?

src/Main/Autowiring.php
```diff
@@ -333,4 +364,47 @@ private function convertDependencyTreeIntoDefinitionList(array $dependencyTree)
 
 		return $classes;
 	}
+
+	/**
+	 * Validates that all classes / interfaces / traits / etc provided here are valid (we can build a ReflectionClass on them
```
dingo-d, line 369:
```suggestion
	 * Validates that all classes
	 *
	 * Validates that all classes/interfaces/traits/etc. provided here are valid (we can build a ReflectionClass on them
```

src/Main/Autowiring.php
```diff
@@ -333,4 +364,47 @@ private function convertDependencyTreeIntoDefinitionList(array $dependencyTree)
...
+	 * ) and return them. Otherwise throw an exception.
+	 *
+	 * @param array $classNames FQCNs found in $this->namespace.
+	 * @param bool  $skipInvalid Skip invalid namespaces rather than throwing an exception. Used for tests.
+	 * @return array
+	 *
+	 * @throws NonPsr4CompliantClass When a found class / file doesn't match PSR-4 standards.
```
dingo-d, line 376:
```suggestion
	 * @throws NonPsr4CompliantClass When a found class/file doesn't match PSR-4 standards.
```

tests/Main/AutowiringTest.php
```diff
@@ -0,0 +1,165 @@
...
+use Tests\Datasets\Autowiring\Dependencies\SubNamespace1\SomeClass;
+use Tests\Datasets\Autowiring\NonServices\SomeFactory;
+use Tests\Datasets\Autowiring\Services\ServiceWithClassDep;
+use Tests\Datasets\Autowiring\Services\ServiceWithDeepClassDep;
+use Tests\Datasets\Autowiring\Services\ServiceWithInterfaceDep;
+use Tests\Datasets\Autowiring\Services\ServiceWithInterfaceDepMoreThanOneClassFound;
+use Tests\Datasets\Autowiring\Services\ServiceWithInterfaceDepWrongName;
+use Tests\Datasets\Autowiring\Services\ServiceWithMultipleDeps;
+use Tests\Datasets\Autowiring\Services\ServiceWithPrimitiveDep;
+use Tests\Datasets\Autowiring\Services\ServiceWithPrimitiveDepHasDefault;
```
dingo-d, line 24:
```suggestion
use Brain\Monkey;
use EightshiftBoilerplate\Main\MainExample;
use EightshiftLibs\Exception\{InvalidAutowireDependency, NonPsr4CompliantClass};
use Tests\Datasets\Autowiring\Deep\Deeper\ServiceNoDependenciesDeep;
use Tests\Datasets\Autowiring\Dependencies\{ClassDepWithNoDependencies,
	ClassImplementingInterfaceDependency,
	ClassWithDependency,
	InterfaceDependency
};
use Tests\Datasets\Autowiring\Dependencies\SubNamespace1\SomeClass;
use Tests\Datasets\Autowiring\NonServices\SomeFactory;
use Tests\Datasets\Autowiring\Services\{ServiceNoDependencies,
    ServiceWithClassDep,
	ServiceWithDeepClassDep,
	ServiceWithInterfaceDep,
	ServiceWithInterfaceDepMoreThanOneClassFound,
	ServiceWithInterfaceDepWrongName,
	ServiceWithMultipleDeps,
	ServiceWithPrimitiveDep,
	ServiceWithPrimitiveDepHasDefault
};
```

dingo-d (changes requested):
Great job! And the tests 🤩

Left a few minor comments for improvements, and one question.

src/Main/Autowiring.php
IvanGrginovInf: @dingo-d  What was changed here is that all reflection classes are built at the top and then passed down (so we don't build the same object in multiple place), nothing more.

As for this method in general, it builds a helper (index) array that is later used to determining which class to inject in a super-fast way (when a class has interface-typehinted dependency) in `private function tryToFindMatchingClass()` . The format of this array is:

```
YourNamespace\SomeClass => [
  YourNamespace\FirstInterfaceThisClassImplements,
  YourNamespace\SecondInterfaceThisClassImplements,
  ...
];
```

Not sure if that answers your question. 😄

tests/Main/AutowiringTest.php
IvanGrginovInf: Huh, I did not know you could do that. And from PHP >=7.0. Nice!

gabriel-glo (approved):

iruzevic (approved):

dingo-d (approved):

PR #178, opened 2021-04-08
updating helper for css variables

iruzevic:
adding options to convert css variable output to color, select-variable, boolean-variable and default. also option to add any css variable no mater the manifest

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -388,7 +388,7 @@ class Components
 			}
 
 			// Output select variable from the options array but dont use value key. It will use variable key.
-			if ($manifest['attributes'][$key]['variable'] === 'select-variable' && isset($manifest['options'][$key])) {
+			if (isset($manifest['options'][$key]) && $manifest['attributes'][$key]['variable'] === 'select-variable') {
 				$selectVariable = array_values(array_filter(
 					$manifest['options'][$key],
 					function ($item) use ($attributes, $key) {
@@ -400,7 +400,7 @@ class Components
 			}
 
 			// Output boolean variable from the options array key. First key is false value, second is true value.
-			if ($manifest['attributes'][$key]['variable'] === 'boolean-variable' && isset($manifest['options'][$key]) && count($manifest['options'][$key]) === 2) {
+			if (isset($manifest['options'][$key]) && $manifest['attributes'][$key]['variable'] === 'boolean-variable' && count($manifest['options'][$key]) === 2) {
 				$value = $manifest['options'][$key][(int)$attributes[$key]];
 			}
 
@@ -415,7 +415,7 @@ class Components
 		}
 
 
-		// Output manual output from the array ov variables.
+		// Output manual output from the array of variables.
 		$manual = isset($manifest['variables']) ? implode(";\n", $manifest['variables']) : '';
 
 		return "
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -382,8 +382,31 @@ class Components
 				continue;
 			}
 
-			if (isset($manifest['attributes'][$key]['color'])) {
+			// Output color variable from the global variables.
+			if ($manifest['attributes'][$key]['variable'] === 'color') {
 				$value = "var(--global-colors-{$value})";
+			}
+
+			// Output select variable from the options array but dont use value key. It will use variable key.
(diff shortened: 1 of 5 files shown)


src/Helpers/Components.php
```diff
@@ -382,19 +382,47 @@ public static function outputCssVariables(array $attributes, array $manifest, st
...
 			}
 
-			if (isset($manifest['attributes'][$key]['color'])) {
+			// Output color variable from the global variables.
+			if ($manifest['attributes'][$key]['variable'] === 'color') {
 				$value = "var(--global-colors-{$value})";
 			}
 
+			// Output select variable from the options array but dont use value key. It will use variable key.
+			if ($manifest['attributes'][$key]['variable'] === 'select-variable' && isset($manifest['options'][$key])) {
```
dingo-d, line 391: I'd put the isset key check first, then the specific value second. Just think it's more readable and logical (checking if it's set first, then the value)

src/Helpers/Components.php
```diff
@@ -382,19 +382,47 @@ public static function outputCssVariables(array $attributes, array $manifest, st
...
+					function ($item) use ($attributes, $key) {
+						return $item['value'] === $attributes[$key];
+					}
+				))[0]['variable'] ?? null;
+
+				$value = $selectVariable === null ? $attributes[$key] : $selectVariable;
+			}
+
+			// Output boolean variable from the options array key. First key is false value, second is true value.
+			if ($manifest['attributes'][$key]['variable'] === 'boolean-variable' && isset($manifest['options'][$key]) && count($manifest['options'][$key]) === 2) {
```
dingo-d, line 403: Same here. Isset first, then other checks.

src/Helpers/Components.php
```diff
@@ -382,19 +382,47 @@ public static function outputCssVariables(array $attributes, array $manifest, st
...
+				$value = $manifest['options'][$key][(int)$attributes[$key]];
+			}
+
+			// Return correct boolean type as string.
+			if (gettype($value) === 'boolean') {
+				$value = $value ? 'true' : 'false';
```
dingo-d, line 409: I think you could just typecast to a string `(string)$value`.

src/Helpers/Components.php
```diff
@@ -382,19 +382,47 @@ public static function outputCssVariables(array $attributes, array $manifest, st
...
+			}
+
 			$key = self::camelToKebabCase($key);
 
 			$output .= "--{$key}: {$value};\n";
 		}
 
+
+		// Output manual output from the array ov variables.
```
dingo-d, line 418:
```suggestion
		// Output manual output from the array of variables.
```

dingo-d (changes requested):
Left some small comments. Awesome that you have tests so you can check if anything breaks after changing the code 😁😁

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -359,12 +359,11 @@ function ($argument) {
 /**
  * Components::outputCssVariables tests
  */
-test('Asserts that outputCssVariables returns the correct css variables output', function () {
+test('Asserts that outputCssVariables returns the correct css variables output for default type', function () {
```
goranalkovic-infinum, line 362:
```suggestion
test('Asserts that outputCssVariables returns the correct CSS variables output for default type', function () {
```

goranalkovic-infinum (changes requested):
Left one suggestion, @dingo-d said the rest 😄  Looks awesome!

src/Helpers/Components.php
volfkarlo, line 409: I'm not fully sure how would do you handle true or false in CSS variable world. 
Boolean in CSS vars, for now, is mostly resolved by 1 or 0 multiplied or by a value that you want to provide, or strings that you pass for example
```
--displaying-variable: block;
--displaying-variable: none;
display: var(--displaying-variable);
```

volfkarlo (changes requested):

src/Helpers/Components.php
iruzevic: I have tried it and it returns 0 and 1 as string.

goranalkovic-infinum (approved):

iruzevic:
Implemented all corrections.

volfkarlo (approved):

PR #181, opened 2021-04-14
updating fixes

iruzevic:
- adding textdomain correct replace
- fixing some warnings in search/replace
- fixing @package rename


src/Cli/AbstractCli.php
```diff
@@ -465,6 +479,22 @@ public function renameUseFrontendLibs(array $args = []): self
 		return $this;
 	}
 
+	/**
+	 * Convert camel case to kebab case.
+	 *
+	 * @param string $input Input to convert.
+	 * @return string
+	 */
+	public function camelCaseToKebabCase(string $input): string
```
IvanGrginovInf, line 488: Unit test for this one por favor 🙏

iruzevic: this is not covered fully in the tests so I will add the tests in the next PR.

Setup is failing now so this is a hotfix

src/Cli/AbstractCli.php
```diff
@@ -465,6 +479,22 @@ public function renameUseFrontendLibs(array $args = []): self
...
+	{
+		preg_match_all('!([A-Z][A-Z0-9]*(?=$|[A-Z][a-z0-9])|[A-Za-z][a-z0-9]+)!', $input, $matches);
```
dingo-d, line 490: What's the use of `!` in the regex here exactly?

Seems to me this helper should either be `private` or put in the helpers as a static helper. And tested ofc.

dingo-d (changes requested):
Left one comment. 

Can you add a test for text-domain replacement as well? So that we verify it works correctly.

IvanGrginovInf (approved):

IvanGrginovInf:
@iruzevic approved but phpstan is failing.

iruzevic:
@IvanGrginovInf fixed phpstan

IvanGrginovInf (approved):

src/Cli/CliHelpers.php
```diff
@@ -37,4 +37,19 @@ public static function cliError(string $errorMessage): void
...
+	 * @param string $input Input to convert.
+	 * @return string
+	 */
+	public static function camelCaseToKebabCase(string $input): string
+	{
+		$output = ltrim(strtolower((string)preg_replace('/[A-Z]([A-Z](?![a-z]))*/', '-$0', $input)), '-');
+		$output = str_replace('_', '-', $output);
+		$output = str_replace('--', '-', $output);
+
+		return $output;
```
dingo-d, line 53:
```suggestion
		return str_replace('--', '-', $output);
```

dingo-d (changes requested):
One small comment left

dingo-d (approved):

PR #182, opened 2021-04-15
Fix faulty test

dingo-d:
One test outputted the contents out and the assertion was faulty so I fixed this.

### DIFF
--- a/tests/Blocks/BlocksExampleTest.php
+++ b/tests/Blocks/BlocksExampleTest.php
@@ -222,9 +222,7 @@ test('Asserts that renderWrapperView will return a valid file.', function () {
 	$this->blocksExample->renderWrapperView($wrapperManifest, []);
 	$content = ob_get_clean();
 
-	$this->assertSame('
-<div>Wrapper!</div>
-', $content);
+	$this->assertSame('<div>Wrapper!</div>', trim($content));
 });
 
 test('Asserts that renderWrapperView will throw error if path is not valid.', function () {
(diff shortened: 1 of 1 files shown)


tests/Blocks/BlocksExampleTest.php
```diff
@@ -218,9 +218,13 @@
...
 	$wrapperManifest = dirname(__FILE__, 2) . '/data/src/Blocks/wrapper/wrapper.php';
 
+	ob_start();
 	$this->blocksExample->renderWrapperView($wrapperManifest, []);
+	$content = ob_get_clean();
 
-	$this->assertTrue(true);
+	$this->assertSame('
+<div>Wrapper!</div>
+', $content);
```
IvanGrginovInf, line 227: How about this:

```suggestion
	$this->assertSame('<div>Wrapper!</div>', trim($content));
```

A BIT less safe but much more pretty.

IvanGrginovInf (approved):
@dingo-d approved but left a suggestion

iobrado (approved):

tests/Blocks/BlocksExampleTest.php
dingo-d: Nope, Doesn't work, because the `renderWrapperView` will actually output the HTML to the OB. So all the break lines will be preserved. 

I added it like this the first time and the test failed 😂

iruzevic (approved):

PR #184, opened 2021-04-19
Fixes incorrect folder detection when used in plugin

### LINKED ISSUE #183: [BUG] - Config::getProjectPath() not working correctly when BP used as a plugin
`Config::getProjectPath()` not working correctly when BP is used as a plugin.

Steps to reproduce:
1. Get a vanilla install with twentytwentyone theme
2. Install new BP plugin
3. `wp boilerplate setup_theme` will fail because of missing global blocks `manifest.json` (because `Config::getProjectPath()` looks for it in `wp-content/themes/twentytwentyone` folder).

Error:
```
Uncaught VirtualGalleryVendor\EightshiftLibs\Exception\InvalidBlock: Global blocks settings manifest.json is missing on this location: /Users/YOUR.USER.NAME/valet-park/site-name/wp-content/themes/twentytwentyone//src/Blocks/manifest.json. in /Users/YOUR.USER.NAME/valet-park/site-name/wp-content/plugins/YOUR-PLUGIN-NAME/vendor/infinum/eightshift-libs/src/Exception/InvalidBlock.php:100 Stack trace: #0 /Users/YOUR.USER.NAME/valet-park/site-name/wp-content/plugins/YOUR-PLUGIN-NAME/vendor/infinum/eightshift-libs/src/Blocks/AbstractBlocks.php(369)
...
```

How to (temporarily fix it).
---
In your plugin's Config.php override parent (`AbstratConfigData`) `getProjectPath()` method like so:
```php
	public static function getProjectPath(string $path = ''): string {
		return \trailingslashit(dirname(dirname(__DIR__)));
	}
```

IvanGrginovInf:
Tested both with plugin and theme. if this generic folder detection isn't defined first, BP plugin will break if you use it with another theme (because it will try to look for BP files inside the active theme folder first)

Changelog
---
* Fixes #183


IvanGrginovInf:
@dingo-d do you perhaps know why tests are failing on this? Localy running `composer standards:check` on PHP 7.4.16 doesn't throw any errors. Plus the error from CI is (it shouldn't fail on those):

```
Error: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '"{$e->getCode()}: {$e->getMessage()}"'.
Error: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found ''<div class="%s">''.
Error: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found ''</div>''.
Error: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '"{$e->getCode()}: {$e->getMessage()}"'.
```

dingo-d:
Have you make `composer update`? 😄  Coding standards have updated, and the escaping is also needed on `exit` constructs (or simply ignoring them would suffice in these cases) 🙂

IvanGrginovInf:
Did `composer update`, not it throws them locally, thx. 👍 

@dingo-d But on another note, shouldn't this one be excluded for example:

```
... found ''</div>''.

echo '</div>';
```

I.e. you're echoing a constant string (no variables in it), no need to escape that right? (that's how it was before)

IvanGrginovInf:
It's also throwing an error on the following:

```
printf('<div class="%s">', \esc_attr("{$attributes['parentClass']}__{$component}"));
```

Which should be fine because the parameter %s is escaped.

dingo-d:
The problem is that there is no easy way to identify what is or isn't escaped inside the `printf` in phpcs, that's why it asks to escape the entire construct I guess.

For the `div` one, that seems like a false positive, not sure why this is happening. I'll try to debug this in the CS

IvanGrginovInf:
> The problem is that there is no easy way to identify what is or isn't escaped inside the printf in phpcs, that's why it asks to escape the entire construct I guess.

This worked on previous versions (for example this is how it behaves on my project on BP 4):
- `printf('<div class="%s">', $class);` - would throw "unescaped output" error.
- `printf('<div class="%s">', esc_attr($class));` - would not

Cause you can't really escape entire `printf` (it echos, doesn't return. We'd have to use `sprintf` and even then you wouldn't be able to escape this properly.)

For example, how do you escape: `sprintf('<div class="%s">', $class);`? If you do `esc_html()` it will break the HTML. If you do `esc_attr()` that's technically incorrect.

dingo-d:
I've just added this to the test in the latest CS and I'm not getting any error, which is super odd. Do we have any kind of other version constraints on this?

IvanGrginovInf:
@dingo-d, @iruzevic please approve this, we need to deal with PHPCS issues in another PR as it's unrelated to thi.

iruzevic (approved):

dingo-d (approved):

PR #185, opened 2021-04-20
Allow infinite number of levels in dependency tree

IvanGrginovInf:
Currently Autowiring didn't work with more than 3 lvls.

This worked:
```
Service
  Lvl1DependencyClass
    Lvl2DependencyClass
```

This did not:
```
Service
  Lvl1DependencyClass
    Lvl2DependencyClass
      Lvl3DependencyClass
```

Now it works infinitely deep by integrating over the `$dependencyTree` array using a pointer. Whenever we find dependencies of a class, the original `$dependencyTree` will be modified and `foreach` will detect those changes and iterate over them without using recursion.

Added tests for Service + 7 lvls of dependencies to confirm this works.


tests/Main/AutowiringTest.php
```diff
@@ -7,15 +7,23 @@
 use Tests\Datasets\Autowiring\Deep\Deeper\ServiceNoDependenciesDeep;
 use Tests\Datasets\Autowiring\Dependencies\{ClassDepWithNoDependencies,
 	ClassImplementingInterfaceDependency,
-	ClassWithDependency,
+    ClassLvl1Dependency,
+    ClassLvl2Dependency,
+    ClassLvl3Dependency,
+    ClassLvl4Dependency,
```
goranalkovic-infinum, line 13: These indents look a bit out of place

goranalkovic-infinum (approved):
Left a comment (a bit of crazy indents), otherwise 🎉  🎉

iruzevic (approved):

tests/Main/AutowiringTest.php
IvanGrginovInf: Not sure why that happened, VSCode automatically added them with 4 spaces (instead of 1 tab) even tho indenting is correctly set for this file. 🤷‍♂️  anyway, fixed

goranalkovic-infinum: If you have the auto-detection of indent style enabled, that can do fun stuff sometimes 😄

dingo-d (approved):

PR #187, opened 2021-04-27
updating variables helper

iruzevic:
- updating css variables helper

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -485,7 +485,7 @@ class Components
 	}
 
 	/**
-	 * Check if provided array is associative or sequential
+	 * Check if provided array is associative or sequential. Will return true if array is sequential.
 	 *
 	 * @param array $array Array to check.
 	 *
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -377,6 +377,8 @@ class Components
 
 		$name = self::camelToKebabCase($name);
 
+		$customOutput = '';
+
 		// Check manifest for the attributes with variable key.
 		$defaultAttributes = array_filter(
 			$manifest['attributes'],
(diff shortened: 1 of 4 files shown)


volfkarlo (approved):
👏

src/Helpers/Components.php
```diff
@@ -472,6 +484,18 @@ public static function camelToKebabCase(string $string): string
...
 	}
 
+	/**
+	 * Check if provided array is associative or sequential
+	 *
+	 * @param array $array Array to check.
+	 *
+	 * @return boolean
+	 */
+	public static function isAssoc(array $array): bool
```
dingo-d, line 494: Have you checked this implementation: https://wiki.php.net/rfc/is_list (the is_list will land in PHP 8.1 from what I see)?

I'm getting different results when I try their version and your helper

https://3v4l.org/M3vNs

goranalkovic-infinum (approved):

tests/Datasets/Arguments.php
```diff
@@ -110,6 +110,20 @@
   ]]
 ]);
 
+dataset('isAssocWrong', [[
+	['a', 'b', 'c'],
+	['1', '2', '3'],
+	[],
+	'',
+	false,
+	['1' => 'a', '0' => 'b', '2' => 'c'],
```
IvanGrginovInf, line 119: This one should still count as assoc array, no?

IvanGrginovInf (approved):
@iruzevic  Approved but left a comment regarding `isAssoc` test

dingo-d (changes requested):
One test is wrong, that Grga mentioned, and I've left a comment about that polyfill, can you check it out?

You can also add the examples from the snippet to the test to see what will happen.

tests/Datasets/Arguments.php
iruzevic: yea this was switched now with the new code that Denis provided

src/Helpers/Components.php
iruzevic: I have updated the new the function that you provided

iruzevic:
@dingo-d not sure whey phpcs fails. I have run it locally and nothing reports

dingo-d:
I am getting some errors, but also some false positives and I'm not sure why (can't reproduce them in the CS package).

The `{$e->getCode()}: {$e->getMessage()}` errors are legit, since you are printing something out, so you can escape those, but the two below (line 129 and 135 in the Components.php) are weird bug.

dingo-d (approved):

iruzevic:
> I am getting some errors, but also some false positives and I'm not sure why (can't reproduce them in the CS package).
> 
> The `{$e->getCode()}: {$e->getMessage()}` errors are legit, since you are printing something out, so you can escape those, but the two below (line 129 and 135 in the Components.php) are weird bug.

I will test this and fix in the next pr

PR #189, opened 2021-04-30
Feature/variables

iruzevic:


dingo-d (approved):

gabriel-glo (approved):

src/Helpers/Components.php
```diff
@@ -431,87 +435,161 @@ function ($item) {
...
+
+		// Output the style for CSS variables.
 		return "
 			<style>
 				.{$name}[data-id='{$unique}'] {
-					{$output}
-					{$manual}
+					{$finalOutput}
 				}
+				{$customResponsiveOutput}
```
IvanGrginovInf, line 567: Is all this being escaped in components / blocks? Because you could in theory do some evil things with this. I.e. if

```
$finalOutput = `</style><script>do evil things</script>...`;
```

iruzevic: I will add additional tests

PR #190, opened 2021-05-03
updating responsive-select

iruzevic:
- the same as on frontend libs.
- escaping outputs

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -447,14 +447,14 @@ class Components
 			$variableType = $manifest['attributes'][$key]['variable'];
 
 			switch ($variableType) {
-				case 'color': {
+				case 'color':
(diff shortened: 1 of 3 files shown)


goranalkovic-infinum (approved):

src/Helpers/Components.php
```diff
@@ -446,13 +447,14 @@ function ($item) {
 			$variableType = $manifest['attributes'][$key]['variable'];
 
 			switch ($variableType) {
-				case 'color':
+				case 'color': {
```
dingo-d: The `case` statement shouldn't have `{` in PHP

iruzevic: this was fixed

dingo-d (approved):

PR #191, opened 2021-05-03
Feature/variables

iruzevic:
- the same as on frontend libs

### DIFF
--- a/src/Blocks/AbstractBlocks.php
+++ b/src/Blocks/AbstractBlocks.php
@@ -353,7 +353,7 @@ abstract class AbstractBlocks implements ServiceInterface, RenderableBlockInterf
 	protected function getComponents(): array
 	{
 
-		$paths = preg_grep('/^([^.])/', scandir($this->getBlocksComponentsPath()) ?? []);
+		$paths = preg_grep('/^([^.])/', (array) scandir($this->getBlocksComponentsPath()) ?? []);
 
 		if (!$paths && !defined('WP_CLI')) {
 			throw InvalidBlock::missingComponentsException();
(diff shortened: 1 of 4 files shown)


src/Blocks/AbstractBlocks.php
```diff
@@ -359,7 +359,7 @@ protected function getComponents(): array
 			throw InvalidBlock::missingComponentsException();
 		}
 
-		$components = array_diff((array)$paths, ['..', '.']);
+		$components = array_diff((array)$paths, ['..', '.', '.DS_Store']);
```
dingo-d, line 362: Why is this added? This is mac specific, and shouldn't be in the code.

iruzevic: I have no idea for some reason 15 test started to fail for no apparent reason, I tried everything.

dingo-d (approved):

dingo-d (changes requested):
DS_Store should be removed.

src/Blocks/AbstractBlocks.php
dingo-d: Because you had `.DS_Store`? That makes no sense.

iruzevic: > DS_Store should be removed.

Done. I have removed all . files from the scan

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -402,377 +389,260 @@ function ($argument) {
...
-	$attributes = [
-		'testSelectVariable' => "test-3",
-	];
-
-	$output = Components::outputCssVariables(
-		$attributes,
-		$manifest,
-		'uniqueString'
-	);
+	var_dump($output);
```
IvanGrginovInf, line 573: Remove var_dump

IvanGrginovInf (changes requested):
Single comment regarding var_dump, otherwise looks good.

dingo-d (approved):

tests/Helpers/ComponentHelpersTest.php
iruzevic: fixed

PR #192, opened 2021-05-05
updating variables

iruzevic:
at this point just don't ask why


goranalkovic-infinum (approved):

volfkarlo (approved):
Seems good. Much shortened code which I like a lot 😄

PR #194, opened 2021-05-17
Updates

dingo-d:
Updated packages
Added new tests

Would still like to see what can be tested and do this before the release

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,15 +28,15 @@
     "ext-dom": "*",
     "ext-json": "*",
     "ext-libxml": "*",
-    "php-di/invoker": "^2.0",
-    "php-di/php-di": "^6.3.0"
+    "php-di/invoker": "^2.3",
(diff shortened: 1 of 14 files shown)


IvanGrginovInf (approved):

iruzevic (approved):

gabriel-glo (approved):

dingo-d:
@infinum/wordpress-team I've fixed the manifest test and added some more. I found out that we needed an additional flag for some functionality to work, which in turn kinda broke the enqueue tests so I've fixed those as well.

Also made some small fixes and added a test for the abstract menu, which now checks for the menu generation (not fully because that would require mocking the entire walker which is just not practical).

Once checked and approved this can be merged :) 

The pcov coverage locally is 88.76% 🙂

IvanGrginovInf (approved):

iruzevic (approved):
looks good

Issue #153, opened 2021-03-04
[BUG] Creating custom post type wrongly sets the label textdomain to name of the CPT


dingo-d:
## Describe the bug
When creating a custom post type using

```bash
wp boilerplate create_post_type
```

The textdomain in the label will be the same as the name of the custom post type

## Steps to Reproduce
1. Setup theme
2. Try to create a custom post type
3. Check the `label` field in the `getPostTypeArguments()` method
4. The text domain shouldn't be the name of the CPT

## Your Environment
* PHP version: 7.4.14
* Development environment: Valet
* Server type: Nginx
* Operating System and version: MacOS Bug Sur 11.2.2
* Libs version: 3.0.4

Issue #152, opened 2021-03-04
[BUG] Creating taxonomy will capitalize the textdomain in the contents


dingo-d:
## Describe the bug
When creating a custom taxonomy using

```bash
wp boilerplate create_taxonomy
```

The textdomain in the labels will be capitalized. Not sure what happens to text domain with a dash in it.

## Steps to Reproduce
1. Setup theme
2. Try to create a custom taxonomy
3. Check the `label` field in the `getTaxonomyArguments()` method
4. The text domain should be lowercased

## Your Environment
* PHP version: 7.4.14
* Development environment: Valet
* Server type: Nginx
* Operating System and version: MacOS Bug Sur 11.2.2
* Libs version: 3.0.4

iruzevic:
Currently textdomain is replaces with the project name space. If you agree I would leave it like this.

I have tested and the bug here was fixed

iruzevic:
we can lower case namespace but if you have something like this 

SomethingCoolNamespace you will get somethingcoolnamespace that looks ugly

dingo-d:
From the [official wp docs](https://developer.wordpress.org/themes/functionality/internationalization/#text-domain)

> The text domain name must use dashes and not underscores and be lowercase. For example, if the theme’s name My Theme is defined in the style.css or it is contained in a folder called my-theme the text domain should be my-theme.

iruzevic:
ok probam vidjeti kaj mogu napraviti

IvanGrginovInf:
> @iruzevic iruzevic linked a pull request that will close this issue on 14 Apr

This seems to work / has been fixed in a previous PR, closing issue.

PR #195, opened 2021-05-19
Minor setup fixes

IvanGrginovInf:
Changelog
---
* Fix phpstan errors that pop up after certain classes / blocks are generated
* Fix broken setup using latest libs (`release/3.1.0`) / frontend libs (`develop`) because of missing `link` component / block
* Fixed a warning during `wp boilerplate create_post_type` if `menu_position` parameter is not provided

#153, #152 (which this PR was supposed to fix) seem to already be fixed, the text domain is properly set, for example this is one I just created in my test theme (same thing for CPT):

```php
	/**
	 * Get the arguments that configure the custom taxonomy.
	 *
	 * @return array Array of arguments.
	 */
	protected function getTaxonomyArguments(): array
	{
		return [
			'hierarchical' => true,
			'label' => \esc_html__('Some Taxonomy', 'boilerplate301'),
			'show_ui' => true,
			'show_admin_column' => true,
			'show_in_nav_menus' => false,
			'public' => true,
			'show_in_rest' => true,
			'query_var' => true,
			'rest_base' => static::REST_API_ENDPOINT_SLUG,
			'rewrite' => [
				'hierarchical' => true,
				'with_front' => false,
			],
		];
	}
```

### DIFF
--- a/src/Blocks/BlocksExample.php
+++ b/src/Blocks/BlocksExample.php
@@ -78,12 +78,12 @@ class BlocksExample extends AbstractBlocks
 	/**
 	 * Set props helper to a proper values to be used in blocks/components.
 	 *
-	 * @param array  $attributes Object of attributes from block/component.
-	 * @param string $realName Old key to use, generally this is the name of the block/component.
-	 * @param string $newName New key to use to rename attributes.
-	 * @param bool   $isBlock Check if helper is used on block or component.
+	 * @param array<mixed> $attributes Object of attributes from block/component.
+	 * @param string       $realName Old key to use, generally this is the name of the block/component.
+	 * @param string       $newName New key to use to rename attributes.
+	 * @param bool         $isBlock Check if helper is used on block or component.
 	 *
-	 * @return array
+	 * @return array<mixed>
 	 */
 	public static function props(array $attributes, string $realName, string $newName = '', bool $isBlock = false): array
 	{
@@ -102,7 +102,7 @@ class BlocksExample extends AbstractBlocks
 			\esc_html__('Blocks', 'eightshift-libs'),
 			self::REUSABLE_BLOCKS_CAPABILITY,
 			'edit.php?post_type=wp_block',
-			'',
+			'', // @phpstan-ignore-line
 			'dashicons-editor-table',
 			4
 		);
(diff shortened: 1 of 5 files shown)


src/CustomPostType/PostTypeCli.php
```diff
@@ -109,7 +109,7 @@ public function __invoke(array $args, array $assocArgs) // phpcs:ignore
 		$rewriteUrl = $this->prepareSlug($assocArgs['rewrite_url']);
 		$restEndpointSlug = $this->prepareSlug($assocArgs['rest_endpoint_slug']);
 		$capability = $assocArgs['capability'] ?? '';
-		$menuPosition = $assocArgs['menu_position'] ? (string)$assocArgs['menu_position'] : '';
+		$menuPosition = isset($assocArgs['menu_position']) ? (string)$assocArgs['menu_position'] : '';
```
goranalkovic-infinum, line 112: Could this be replaced with
```php
$menuPosition = (string) ($assocArgs['menu_position'] ?? '');
```

IvanGrginovInf: yup, done!

src/Blocks/BlocksExample.php
```diff
@@ -78,12 +78,12 @@ protected function getBlocksPath(): string
...
 	 * Set props helper to a proper values to be used in blocks/components.
 	 *
-	 * @param array  $attributes Object of attributes from block/component.
-	 * @param string $realName Old key to use, generally this is the name of the block/component.
-	 * @param string $newName New key to use to rename attributes.
-	 * @param bool   $isBlock Check if helper is used on block or component.
+	 * @param array<mixed> $attributes Object of attributes from block/component.
+	 * @param string       $realName Old key to use, generally this is the name of the block/component.
+	 * @param string       $newName New key to use to rename attributes.
+	 * @param bool         $isBlock Check if helper is used on block or component.
```
dingo-d, line 84: If memory serves me, this is no longer needed (indenting so it's aligned with doc block comments).

dingo-d (approved):

iruzevic (approved):

src/Blocks/BlocksExample.php
IvanGrginovInf: Not needed but all functions seem to still have this. Dunno I'd kinda prefer this but we can remove this everywhere as well. 🤷‍♂️

PR #196, opened 2021-05-20
Kebab-case to camel case helper

goranalkovic-infinum:
Adds a kebab-to-camelCase helper (and a corresponding test) for cases when you need camelCased things (e.g. using the `Blocks::props` with a component that has a `-` in the name).

### DIFF
--- a/tests/Helpers/ComponentHelpersTest.php
+++ b/tests/Helpers/ComponentHelpersTest.php
@@ -754,6 +754,33 @@ test('Asserts that kebabToCamelCase returns the wrong output', function () {
 });
 
 /**
+ * Components::kebabToCamelCase tests
+ */
+test('Asserts that kebabToCamelCase returns the correct output with a different separator', function () {
+	$output = Components::kebabToCamelCase('super_cool_test_string', '_');
+
+	$this->assertEquals('superCoolTestString', $output);
+});
+
+/**
+ * Components::kebabToCamelCase tests
+ */
+test('Asserts that kebabToCamelCase returns the correct output with numbers as a string', function () {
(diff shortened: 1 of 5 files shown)


gabriel-glo (approved):

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -738,6 +738,21 @@ function ($argument) {
...
+
+	$this->assertEquals('superCoolTestString', $output);
+});
+
+test('Asserts that kebabToCamelCase returns the wrong output', function () {
+	$output = Components::kebabToCamelCase('super-cool-test-string-goc');
+
+	$this->assertNotEquals('super_CoolTest-String goc', $output);
+});
+
```
IvanGrginovInf, line 755: Can you try adding a few more tests? Such as what happens when you pass it a non-kekbab-case string? Numbers as string? Or a a different separator (since that's an optional argument not tested here)?

IvanGrginovInf (changes requested):
Left a comment regarding adding more tests, other than that good job!

tests/Helpers/ComponentHelpersTest.php
iruzevic: Yes check out my data set for camelToKebabCase helper

goranalkovic-infinum: I added datasets for correct and wrong examples (based on the camelToKebabCase ones), added the function to the CLIHelpers and wrote additional tests to use the datasets.

volfkarlo (approved):

PR #198, opened 2021-05-25
Register BlockPattern CLI command

IvanGrginovInf:


src/Cli/Cli.php
```diff
@@ -87,6 +87,7 @@ class Cli
 		BlockComponentCli::class,
 		BlockWrapperCli::class,
 		BlockVariationCli::class,
+		BlockPatternCli::class,
```
dingo-d, line 90: Why is this erroring out? Also, I think that one unit test is failing...

dingo-d (changes requested):
Can you address the failing tests and issues?

IvanGrginovInf:
We have 2 tests which check the exact number of classes in CLI. Meaning when you add a new class the test will automatically fail which I'm not too happy about. Regardless, it's fixed and the tests should pass now.

dingo-d (approved):

iruzevic (approved):

PR #200, opened 2021-05-26
Fix plugin install for GitHub plugins

IvanGrginovInf:
Because of the way `wp plugin install` works with remote zips (it appends a random hash at the end of the file after unzipping) and you can't change the name of the archive, we need to do it like this:

1. Download `release.zip` from a repo and save it as the correct plugin name (for example: `eightshift-forms.zip`)
2. Install from the local file and delete afterwards

### DIFF
--- a/src/Cli/CliHelpers.php
+++ b/src/Cli/CliHelpers.php
@@ -63,4 +63,23 @@ trait CliHelpers
 	{
 		return lcfirst(str_replace($separator, '', ucwords(mb_strtolower($string), $separator)));
 	}
+
+	/**
+	 * Returns the name of the github plugin without slashes. For example converts "infinum/eightshift-forms" to "eightshift-forms"
+	 *
+	 * @param string $name Name of the github package.
+	 * @return string
+	 */
+	public static function getGithubPluginName(string $name): string
+	{
+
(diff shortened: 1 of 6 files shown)


iruzevic (approved):

dingo-d:
Not sure I understand this. The `wp plugin install https://github.com/{$name}/releases/download/{$version}/release.zip` command won't install a plugin or?

IvanGrginovInf:
> Not sure I understand this. The `wp plugin install https://github.com/{$name}/releases/download/{$version}/release.zip` command won't install a plugin or?

It will install it as `release-asdabdd`. And then the next time you run the update, it will add another plugin with a different name (meaning you will have a duplicate plugin in the list of installed plugins in WP admin) rather than updating the existing one. Also the folder name will be incorrect.

dingo-d:
Wow, didn't know about that. That's odd.

dingo-d (approved):

IvanGrginovInf:
> Wow, didn't know about that. That's odd.

@dingo-d  Yeah, not sure what's happening there. You can try running this in any project 2-3 times and you'll see what I mean.

`wp plugin install https://github.com/infinum/eightshift-forms/releases/download/0.2.2/release.zip`

PR #204, opened 2021-06-08
Responsive variables

volfkarlo:
Changelog:
- added responsive variables handler and its helper functions

PHP part of https://github.com/infinum/eightshift-frontend-libs/pull/362
Works basically the same.

Functions are named the same and function basically the same.

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -398,13 +398,22 @@ class Components
 	 *
 	 * @return array
 	 */
-	private static function setBreakpointResponsiveVariables(array $attributeVariables, string $breakpointName, int $breakpointIndex, int $numberOfBreakpoints): array
-	{
+	private static function setBreakpointResponsiveVariables(
+		array $attributeVariables,
+		string $breakpointName,
+		int $breakpointIndex,
+		int $numberOfBreakpoints
+	): array {
 		$breakpointAttributeValues = [];
 		foreach ($attributeVariables as $attributeVariablesObject) {
-			// Calculate default breakpoint index.
+
+			/**
+			 * Calculate default breakpoint index based on order of the breakpoint, inverse property
+			 * and number of properties in responsiveAttributeObject.
+			 */
 			$defaultbreakpointIndex = $attributeVariablesObject['inverse'] ? 0 : ((int) $numberOfBreakpoints - 1);
 
+			// Expanding an object with an additional breakpoint property.
 			$attributeVariablesObject['breakpoint'] = ($breakpointIndex === $defaultbreakpointIndex) ? 'default' : $breakpointName;
 			$breakpointAttributeValues[] = $attributeVariablesObject;
 		};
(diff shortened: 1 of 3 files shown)


goranalkovic-infinum (approved):

volfkarlo:
I apologize for the request, I didn't write tests yet 🙂

dingo-d (approved):

src/Helpers/Components.php
```diff
@@ -388,6 +388,164 @@ public static function globalInner(array $itemValues, string $itemKey): string
...
+			$breakpointAttributeValues[] = $attributeVariablesObject;
+		};
+
+		return $breakpointAttributeValues;
+	}
+
+	/**
+	 * Iterating through variables matching the keys from responsiveAttributes and translating it to responsive attributes names.
+	 *
+	 * @param array $responsiveAttributes Responsive attributes that are read from component's/block's manifest.
```
IvanGrginovInf, line 426: You should write what types of values are supported here otherwise `phpstan` would complain. Try running `composer analyze` in libs.

Something like:

```suggestion
	 * @param array<array<string>> $responsiveAttributes Responsive attributes that are read from component's/block's manifest.
```

If there's too many variations you can also write `array<mixed>` but ideally this should be used as little as possible.

This applies to all other `array` params / returns inside doc-blocks as well.

volfkarlo: Quite useful info. I'll definitely check it out and make changes.
Thx! 😄

volfkarlo: What I could notice in our libs is that something like `array<...>` occurs 10 times and `@param array` around 100ish times and `@return array` 170 times. I believe the reason behind it is that the structure is waaay more complex than one layer(`array<array<array<array<mixed>>>>`). I would agree that this would be the reason more to do so, but I think that it might be ok not to "crowd" code with it because it's really complex and, what I believe, not a lot of people would read it since you anyway get lost in all of this nesting.

Also sometimes, this kind of type determination is not even possible since our manifest contains different nesting depths. For example, `componentName` is one level deep, and variables field is 4-5 levels deep which again brings so many possibilities and options that, in my opinion, documenting it doesn't bring that much.

> Still, it would be good to add those if possible or if you have/expect consistent types inside array values.
Is it ok if I just leave it like this for now? I found one spot where I could put it and it would be of a type `array<mixed>` which is not that useful.

tests/data/src/Blocks/components/responsiveVariables/manifest.json
```diff
@@ -0,0 +1,122 @@
+{
```
gabriel-glo, line 1: Can this really be classified as a component. To me this would work better as a standalone JSON file in Blocks directory (alongside the main manifest.json file)

iruzevic: this works only on blocks and components so we should follow that folder convention so for me this looks ok

tests/data/src/Blocks/components/responsiveVariables/manifest.json
```diff
@@ -0,0 +1,122 @@
...
+	"componentName": "responsiveVariables",
+	"title": "Responsive Variables",
+	"componentClass": "responsiveVariables",
+	"example": {
+	},
+	"attributes": {
+    "variableValueDesktop": {
```
iruzevic, line 8: please fix indentations in this file

iruzevic (changes requested):

iruzevic (approved):

PR #205, opened 2021-06-15
Output attributes which can be set as 0

IvanGrginovInf:
This fixes a bug where some attributes (like wrapper spacing attributes) won't be outputted as CSS variables if they're set to 0.

This is especially problematic if you have something like the following structure in editor:

```
group         // wrapper-spacing-top == 20
    heading
    paragraph // wrapper-spacing-top == 0
```

Because `paragraph` will inherit the `wrapper-spacing-top: 20` css variable from group block.

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -781,7 +781,7 @@ class Components
 			// If value contains magic variable swap that variable with original attribute value.
 			if (strpos($variableValue, '%value%') !== false) {
 				// Bailout if magic variable is empty.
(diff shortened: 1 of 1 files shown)


dingo-d (approved):

volfkarlo (approved):

dingo-d (approved):

iruzevic (approved):

iruzevic:
Please make the same correction on the JS side to have the behaviour in sync

PR #206, opened 2021-06-18
Props helper fix

goranalkovic-infinum:
Solves a major bug with output order.

If your component had
```
...
componentWhichIsAliasToParagraph
paragraph
...
```
in the dependency list, and you called
```php
Blocks::props($attributes, 'paragraph', 'componentWhichIsAliasToParagraph')
```
, the attributes you want (here belonging to `componentWhichIsAliasToParagraph`) would get overwritten in the next loop (here with the default `paragraph` options) and the output would be wrong.

This PR makes it so you first collect all the attributes from dependencies and then at the end it goes through the component that you want (here `componentWhichIsAliasToParagraph`).

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -859,11 +859,10 @@ class Components
 	 */
 	public static function props(array $attributes, string $realName, string $newName = '', bool $isBlock = false, array $globalData): array // phpcs:ignore PEAR.Functions.ValidDefaultValue.NotAtEnd
 	{
-
 		$newNameInternal = $newName;
 
 		// Check if newName key is passed if not use the default one from block/component name.
-		if (!$newName) {
+		if (empty($newName)) {
 			$newNameInternal = $realName;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Helpers/Components.php
```diff
@@ -905,25 +904,29 @@ function ($item) use ($realName) {
 			$output['componentName'] = $parent;
 		}
 
-		// Loop attributes.
-		foreach ($attributes as $key => $value) {
-			$result = false;
-			foreach ($dependency as $element) {
-				if ($element === substr($key, 0, strlen($element))) {
-					$result =  true;
+		// Loop all attributes but the actual (newNameInternal) component.
```
volfkarlo, line 907: Maybe a tiny bit bigger clarification would be awesome.
More to explain what is it for than what it doing.. For example.

"Searching for the newNameInternal component" or "Extracting something from something else"

src/Helpers/Components.php
```diff
@@ -905,25 +904,29 @@ function ($item) use ($realName) {
...
+		);
 
-			// Check if attributes key exists in the dependency by comparing the keys partial string.
-			if ($result) {
-				$newKey = $key;
-
-				// Change the name of the key if they are different.
-				if ($realName !== $newNameInternal) {
-					$newKey = str_replace($newNameInternal, $realName, $key);
+		foreach ($attributes as $key => $value) {
```
volfkarlo, line 917: It would be awesome if you would be able to name it somehow other than key and value(preName => postName or something)

src/Helpers/Components.php
```diff
@@ -905,25 +904,29 @@ function ($item) use ($realName) {
...
+			foreach ($filteredDependencies as $element) {
+				if (strpos($key, $element) !== false) {
+					$output[$key] = $value;
 				}
+			}
+		}
 
-				// Populate output with new values.
+		// Loop attributes of the actual (newNameInternal) component.
+		foreach ($attributes as $key => $value) {
```
volfkarlo, line 926: Same as above.

volfkarlo (commented):
Left few comments. Mostly to explain functions/variables a bit more in detail.

src/Helpers/Components.php
```diff
@@ -859,11 +859,10 @@ public static function arrayIsList(array $array): bool
 	 */
 	public static function props(array $attributes, string $realName, string $newName = '', bool $isBlock = false, array $globalData): array // phpcs:ignore PEAR.Functions.ValidDefaultValue.NotAtEnd
 	{
-
 		$newNameInternal = $newName;
 
 		// Check if newName key is passed if not use the default one from block/component name.
-		if (!$newName) {
+		if (empty($newName)) {
```
dingo-d, line 865: The above will work. The `$newName` must be a `string`, and it defaults to an empty string `''`, which is a falsy value so `!$newName` will give the same result as `empty($newName)`, without the overhead of calling a function 🙂

dingo-d (changes requested):
One comment left.

dingo-d (approved):

iruzevic (commented):
Is this there the same issue with the frontend libs or this is only here?

volfkarlo (approved):
Maybe not the clearest version of comments. But I can't suggest any better 😄 
Well done Goc!

gabriel-glo (approved):

IvanGrginovInf (approved):

goranalkovic-infinum:
@iruzevic this probably isn't relevant anymore as the whole props helper was rewritten, right?

IvanGrginovInf:
> @iruzevic @volfkarlo @IvanGrginovInf this probably isn't relevant anymore as the whole props helper was rewritten, right?

Yup, the props helper is now completely different.

goranalkovic-infinum:
Closing as solution is not relevant anymore due to current props helper changes.

PR #224, opened 2021-07-24
Additional phpstan fixes

dingo-d:
Fixed some phpstan issues so that the blocks don't scream too much - they'll still scream because of some false positives and how some code is written (especially the `checkAttr` helper that returns everything so it's next to impossible to ensure type safety 😬).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "source": "https://github.com/infinum/eightshift-libs"
   },
   "require": {
-    "php": "^7.3",
+    "php": "^7.3 || <8.0",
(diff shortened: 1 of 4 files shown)


iruzevic (approved):

dingo-d:
https://github.com/phpstan/phpstan/issues/5374

There is a fix for one of the issues phpstan throws. We should use `is_` functions instead of typechecks.

iruzevic (approved):

gabriel-glo (approved):

iruzevic (approved):

PR #227, opened 2021-07-28
Fix variable type errors in Components::variablesInner on PHP 8  

mbmjertan:
The `Components::variablesInner` function is defined with a mixed type parameter `$attributeValue` which is then passed to `str_replace` as a parameter. 

Since PHP 8, [types for internal functions](https://externals.io/message/106522) have been added, including for [str_replace](https://www.php.net/str_replace), so passing anything other than a `string|array` to str_replace throws a fatal error.

As it's sometimes needed to pass an int to variablesInner (for example, in the columns block in https://github.com/infinum/eightshift-frontend-libs/pull/414), it makes sense to cast this parameter to string.

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -865,7 +865,7 @@ class Components
 
 			// If value contains magic variable swap that variable with original attribute value.
 			if (strpos($variableValue, '%value%') !== false) {
-				$variableValue = str_replace('%value%', $attributeValue, $variableValue);
+				$variableValue = str_replace('%value%', (string)$attributeValue, $variableValue);
 			}
 
 			// Output the custom CSS variable by adding the attribute key + custom object key.


IvanGrginovInf (approved):

goranalkovic-infinum (approved):

iruzevic (approved):

dingo-d:
I wouldn't merge this just yet. There are a lot of `str_replace` calls in the lib, and we are not (yet) supporting PHP 8.0.

So we should 

1. Check all the code for PHP 8 compatibility
2. Check if typecasting this variable to string will break anything else (write test cases for this)
3. Find other `str_replace` calls in the lib and see the attribute types

Overall great catch!

dingo-d (changes requested):
Waiting on PHP 8 compatibility

mbmjertan:
> Find other str_replace calls in the lib and see the attribute types

I've taken the liberty of looking at other `str_replace` [calls in the lib](https://github.com/infinum/eightshift-libs/search?q=str_replace), and most the calls call `str_replace` with parameters that are either a result of a function that returns a string or a hard-coded string. 

The only other potentially problematic `str_replace` calls I could find are in `__invoke` methods of CLI classes, where `str_replace` is sometimes called with certain parameters from an array without a set member type.

For instance, in PostTypeCli.php we have:

https://github.com/infinum/eightshift-libs/blob/17e455772f3b4bd13ae460b325a68539c619c1a3/src/CustomPostType/PostTypeCli.php#L107

https://github.com/infinum/eightshift-libs/blob/17e455772f3b4bd13ae460b325a68539c619c1a3/src/CustomPostType/PostTypeCli.php#L135

https://github.com/infinum/eightshift-libs/blob/17e455772f3b4bd13ae460b325a68539c619c1a3/src/CustomPostType/PostTypeCli.php#L28-L46

For instance, if `$args[1]` was set to `(int) 4` when calling `getDevelopArgs` (as was the case here) the same thing would happen. 

As we can't typehint an array, I can think of two ways to ensure only strings are passed to str_replace even if WPCLI passes something other than a string to this method (apart from due diligence) - declaring the return type for `getDevelopArgs` in the DocBlock to `array<string>` so these issues can be identified by static analysis tools, or casting all `$args` members to strings (which doesn't seem like a great solution). Either way, this seems like a minor issue compared to this.

> Check if typecasting this variable to string will break anything else (write test cases for this)
Sure, tests are a good idea for this. However, this casting should have any side effects as the variable is typecast locally, solely as a `str_replace` parameter. The return value is, without a doubt, a string - before and after this change. This change just makes the casting explicit rather than a consequence of type juggling.

> Check all the code for PHP 8 compatibility
Absolutely.

dingo-d:
> As we can't typehint an array...

Sure you can: https://phpstan.org/writing-php-code/phpdoc-types#general-arrays

😄

PR #229, opened 2021-07-29
Fix check attr responsive phpstan issues

IvanGrginovInf:
Same fix as https://github.com/infinum/eightshift-libs/pull/228 but for `checkAttrResponsive`


dingo-d (approved):
😢

iruzevic (approved):

PR #231, opened 2021-08-13
Added undefined attribute option to checkAttr and checkAttrResponsive helpers

volfkarlo:
Changelog:
- added undefined attribute option to checkAttr and checkAttrResponsive helper

Note:
I tried to detect whether the value of a responsive attribute is false or not set(what we need from inheriting) and it turned out I can't distinguish that one with checkAttr nor checkAttrResponsive helpers.

So I figured this could be a reasonable approach. Even though it's technically different from the JS implementation(JS makes difference between `undefined` and `null`). I believe setting parameters to `NULL` makes perfect sense in PHP part.

The reasoning is that in PHP when a variable is unset or not set it is represented as "NULL" if you try to make something with it(even if you check the variable with a function `isset` result is `false`). That's why I believe that setting values to `NULL` when they should represent the JS' "undefined" should be ok.

### DIFF
--- a/tests/Helpers/ComponentHelpersTest.php
+++ b/tests/Helpers/ComponentHelpersTest.php
@@ -168,7 +168,7 @@ test('Asserts that checkAttr works in case attribute is boolean', function () {
 
 	$results = Components::checkAttr('buttonIsAnchor', $attributes, $manifest);
 
-	$this->assertIsBool($results, 'THe result should be a boolean');
+	$this->assertIsBool($results, 'The result should be a boolean');
 	$this->assertSame(true, $results, "The set attribute should be {$attributes['buttonIsAnchor']}");
 });
 
@@ -178,7 +178,7 @@ test('Asserts that checkAttr returns false in case attribute is boolean and defa
 
 	$results = Components::checkAttr('buttonIsNewTab', $attributes, $manifest);
 
-	$this->assertIsBool($results, 'THe result should be a boolean');
+	$this->assertIsBool($results, 'The result should be a boolean');
 	$this->assertSame(false, $results, "The set attribute should be false");
 });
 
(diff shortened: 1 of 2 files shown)


dingo-d:
Can you write tests for this feature? We should check if this null option is being passed in any way to PHP internal functions because that [could bite us in the ass in the future](https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation).

volfkarlo:
@dingo-d If I'm not wrong it shouldn't make any problems. What we have here is a return of a value of a variable. If the variable is not set it would anyway take a value of `null`.

What I can do, if you agree, is to clean the array from 'null' before returning `checkAttrResponsive` array. That would, as it seems to me, be the only place where something can go wrong. But on the other hand, maybe it is better to return "the full" array as it is happening in JS.

goranalkovic-infinum (approved):

iruzevic (approved):

tests/Helpers/ComponentHelpersTest.php
```diff
@@ -182,6 +182,16 @@ function ($argument) {
 	$this->assertSame(false, $results, "The set attribute should be false");
 });
 
+test('Asserts that checkAttr returns null in case attribute is boolean, default is not set and undefined is allowed', function () {
+	$manifest = Components::getManifest(dirname(__FILE__, 2) . '/data/src/Blocks/components/button/');
+	$attributes['buttonIsAnchor'] = true;
+
+	$results = Components::checkAttr('buttonIsNewTab', $attributes, $manifest, true);
+
+	$this->assertIsNotBool($results, 'THe result should not be a boolean');
```
dingo-d, line 191:
```suggestion
	$this->assertIsNotBool($results, 'The result should not be a boolean');
```

dingo-d (approved):
Left one comment, but looks good.

Always good to have tests just to make sure we cover all the edge cases :)

PR #235, opened 2021-09-13
Convert variable to string when replacing value

simun-ivanac:
When looping through css variables, some values are passed as integers. Using `str_replace()` in internal helper `variablesInner` throws fatal error on integers. Convert it to string when replacing value.

### DIFF
--- a/src/Helpers/Components.php
+++ b/src/Helpers/Components.php
@@ -872,7 +872,7 @@ class Components
 
 			// If value contains magic variable swap that variable with original attribute value.
 			if (strpos($variableValue, '%value%') !== false) {
-				$variableValue = str_replace('%value%', $attributeValue, $variableValue);
+				$variableValue = str_replace('%value%', (string) $attributeValue, $variableValue);
 			}
 
 			// Output the custom CSS variable by adding the attribute key + custom object key.


goranalkovic-infinum (approved):

iruzevic (approved):

mbmjertan (approved):
This is the same change as https://github.com/infinum/eightshift-libs/pull/227.

At the time, we (@dingo-d and myself) had a discussion about whether or not to merge this, and decided to wait for PHP 8 compatibility as a whole, as this is only an issue on PHP 8. See https://wiki.php.net/rfc/consistent_type_errors for more info.

I'm still very much for this fix: it doesn't actually change anything - the variable would be cast as a string by type juggling in earlier PHP versions, and here that casting operation is just explicitly written out.

However, writing tests (which Denis suggested) for this seems like a good idea as the `variablesInner` helper isn't covered by tests as far as I can see. (Although, tests for `outputCssVariables` would probably fail if this was breaking anything, as we have a call chain of `outputCssVariables` -> `setVariablesToBreakpoints` ->  `variablesInner`. It doesn't seem as a wise choice to rely on that, though.)

PR #247, opened 2021-12-05
A11y improvements

mbmjertan:
This PR fixes these issues reported in https://github.com/infinum/eightshift-frontend-libs/issues/491:

- [x] "The type attribute is unnecessary for JavaScript resources."
- [x] "The type attribute for the style element is not needed and should be omitted."
- [x] "Element head is missing a required instance of child element title."


volfkarlo (approved):

dadadavorin (approved):
👍

PR #263, opened 2022-03-07
Feature/inline style

iruzevic:
# Description

- Added everything to a global variable $esBlocks just like in JS window object
- refactored output css variable to one style tag
- refactoring getManifest to determine the component/block from path
- changing some function to private
- optimized loading of block for only one read for manifest.json

# Screenshots / Vide
Screenshot 2022-03-07 at 15 15 23
Screenshot 2022-03-07 at 15 15 52
os

Screenshot 2022-03-07 at 15 22 34


src/Blocks/AbstractBlocks.php
```diff
@@ -63,46 +63,44 @@ public function addThemeSupport(): void
...
-			return [];
+		$namespace = $settings['namespace'];
+
+		// Save namespace to internal variable.
+		$this->namespace = $namespace;
+
+		// If namespace is missing bailout.
+		if (isset($esBlocks[$namespace])) {
+			return;
 		}
```
dingo-d, line 79: I'd move this above the line where you save the namespace to an internal variable.

src/Blocks/AbstractBlocks.php
```diff
@@ -63,46 +63,44 @@ public function addThemeSupport(): void
...
 
-		return $key ? $this->blocks[$key] : $this->blocks;
+		$blocks = array_map(
+			function ($block) use ($namespace) {
```
dingo-d, line 82:
```suggestion
			static function ($block) use ($namespace) {
```

src/Helpers/Components.php
```diff
@@ -1043,4 +817,494 @@ function ($a) use (&$return) {
...
+			return $items;
+		}
+
+		// If searching for one item.
+		if ($item) {
+			$items = $details[$type][$item] ?? [];
+
+			if (!isset($details[$type][$item])) {
+				throw InvalidBlock::missingSettingsKeyException($type, $item);
+			}
```
dingo-d, line 877: The `$items` should be placed after the `!isset` check.

src/Blocks/AbstractBlocks.php
```diff
@@ -135,7 +134,7 @@ public function getAllBlocksList($allowedBlockTypes, \WP_Block_Editor_Context $b
 			function ($block) {
 				return $block['blockFullName'];
 			},
-			$this->blocks['blocks'] ?? []
+			Components::getSettings('blocks', '', $this->namespace) ?? []
```
dingo-d, line 137: The return type of `getSettings` is mixed, so it can be `null`?

Because if it cannot be null, the null coalesce operator will never default to `[]`...

src/Blocks/AbstractBlocks.php
```diff
@@ -367,26 +367,32 @@ public function filterBlocksContent(array $parsedBlock, array $sourceBlock): arr
 	 *
 	 * @return string
 	 */
-	abstract protected function getBlocksPath(): string;
+	private function getBlocksPath(): string
+	{
+		$separator = DIRECTORY_SEPARATOR;
```
dingo-d, line 372: I see this used a lot, can we extract this to a private member variable? Class constants won't help, because they cannot be referenced during interpolation.

src/Blocks/AbstractBlocks.php
```diff
@@ -641,6 +653,7 @@ protected function prepareComponentAttributes(array $manifest, string $parent =
 	 */
 	private function getBlocksData(): array
 	{
+		$separator = DIRECTORY_SEPARATOR;
 		return array_map(
 			function (string $blockPath) {
 				$block = implode(' ', (array)file(($blockPath)));
```
dingo-d, line 659: In the case the `$blockPath` is false, the `$block` will be an empty string. The `parseManifest` handles this correctly?

The same is for the 697 line

src/Blocks/AbstractBlocks.php
```diff
@@ -665,7 +678,33 @@ function (string $blockPath) {
...
+	{
+		$separator = DIRECTORY_SEPARATOR;
+		return array_map(
+			function (string $componentPath) {
+				$component = implode(' ', (array)file(($componentPath)));
+
+				$component = $this->parseManifest($component);
+
+				if (!isset($component['componentName'])) {
+					throw InvalidBlock::missingComponentNameException($componentPath);
```
dingo-d, line 702: Tbh, not sure what will happen if you throw an error inside array_map 🙈 We should cover this with tests

https://3v4l.org/P8CQK

dingo-d (changes requested):
Left some comments, tests will need to be updated.

src/Blocks/AbstractBlocks.php
iruzevic: this is old code and it should be covered with test so I think it is ok

src/Blocks/AbstractBlocks.php
iruzevic: agreed

src/Helpers/Components.php
```diff
@@ -138,52 +148,52 @@ public static function render(string $component, array $attributes = [], string
...
-	 * @param array<string, mixed> $attributes Block/Component rendered attributes data.
+	 * @param string $path Absolute path to manifest folder.
+	 * @param bool $useGlobal Use global blocks settings.
+	 *
+	 * @throws ComponentException When we're unable to find the component by $component.
 	 *
 	 * @return array<string, mixed>
 	 */
-	public static function getDefaultRenderAttributes(array $manifest, array $attributes): array
+	public static function getManifest(string $path, $useGlobal = true): array
```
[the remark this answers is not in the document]
iruzevic: I have updated this a bit. getSettings can only return string/array. So it should be ok. I will cover this with test later on

src/Blocks/AbstractBlocks.php
dingo-d: Yeah, if it'll never return null, then you can also remove the `?? []` part.

https://3v4l.org/Zp8Hp

src/Blocks/AbstractBlocks.php
```diff
@@ -28,6 +28,13 @@ abstract class AbstractBlocks implements ServiceInterface, RenderableBlockInterf
 	 */
 	protected $namespace = '';
 
+	/**
+	 * Directory separator used for cross compatibility.
```
dingo-d, line 32:
```suggestion
	 * Directory separator used for OS cross-compatibility.
```

tests/Blocks/BlocksExampleTest.php
```diff
@@ -328,15 +273,15 @@
 
 	$this->blocksExample->registerBlocks();
 
-	$this->assertSame(getenv('BLOCK_TYPE'), 'true', 'Calling void method register_block_type caused no sideaffects');
+	$this->assertSame(getenv('BLOCK_TYPE'), 'true', 'Calling void method register_block_type caused no side affects');
```
volfkarlo, line 276:
```suggestion
	$this->assertSame(getenv('BLOCK_TYPE'), 'true', 'Calling void method register_block_type caused no side effects');
```
I think, not sure.

src/Helpers/Components.php
```diff
@@ -1043,4 +890,483 @@ function ($a) use (&$return) {
...
+	/**
+	 * Extracting the names of default breakpoints depending on the case used in responsive(mobile first/desktop first).
+	 * Returning the 'min' key with default name for mobile first, and the 'max' key for desktop first version.
+	 * If there are no breakpoints, min and max will be empty strings.
+	 *
+	 * @param array<string, mixed> $breakpoints Attributes that are read from component's/block's manifest.
+	 *
+	 * @return array<string, mixed> Associative array with min and max keys.
+	 */
+	private static function getDefaultBreakpoints(array $breakpoints): array
```
volfkarlo, line 1128: Are these breakpoints presorted when entering the function? If they are coming straight from the manifest, I would suggest ensuring it by sorting it by the value of the breakpoint.

volfkarlo (commented):
Just a few questions left 😄 
Awesome job! 😄

src/Helpers/Components.php
```diff
@@ -11,12 +11,27 @@
...
+	public static $namespace = '';
+
+	/**
+	 * Directory separator used for cross compatibility.
+	 *
+	 * @var string
+	 */
+	private static $sep = DIRECTORY_SEPARATOR;
+
 	/**
```
mbmjertan, line 35: Regarding code style here: As this is mostly used to support string interpolation, e.g. in `$sep = self::$sep`and afterwards `$componentPath = "{$parentPath}{$sep}$component"`, I'm not too sure that this should be a class property.

I'd prefer:
1. using `$sep = DIRECTORY_SEPARATOR;` in methods themselves, as we're instantiating method-scope variables either way
2. `$sep` being a class constant (not a property, and not a static property), but at that point we're just adding a class constant that's equal to a global constant

src/Helpers/Components.php
```diff
@@ -1043,4 +890,483 @@ function ($a) use (&$return) {
...
+
+				// If breakpoint is not set or if breakpoint is a default breakpoint use default name.
+				$isDefaultBreakpoint = empty($breakpointItem['breakpoint']) || $breakpointItem['breakpoint'] === $defaultBreakpoints[$type];
+				$breakpoint = $isDefaultBreakpoint ? 'default' : $breakpointItem['breakpoint'];
+
+
+				// Iterate each data array to find the correct breakpoint.
+				foreach ($data as $index => $item) {
+					// Check if breakpoint and type match.
+					if ($item['name'] === $breakpoint && $item['type'] === $type && (!empty((string) $attributeValue) || gettype($attributeValue) === 'integer' || gettype($attributeValue) === 'float' || gettype($attributeValue) === 'double' || $attributeValue === '0')) { // phpcs:ignore Generic.Files.LineLength.TooLong
```
mbmjertan, line 1185: Maybe break these compound conditionals into newlines for readability, e.g. something like this, but with better formatting
```suggestion
					if ($item['name'] === $breakpoint 
					&& $item['type'] === $type 
					&& (!empty((string) $attributeValue) 
						|| gettype($attributeValue) === 'integer' 
						|| gettype($attributeValue) === 'float' 
						|| gettype($attributeValue) === 'double' 
						|| $attributeValue === '0')
					) { 
```

mbmjertan (commented):
This is great!

I've left two comments regarding code style. 

I'll take a closer look at the logic later, as this seems like magic at the moment and it's very difficult to read through the diffs due to the amount of changes, so I'll have to clone it to read through. 👀

src/Blocks/AbstractBlocks.php
mbmjertan: This should have test coverage, yeah, but not sure what you meant by this, @dingo-d:

>  not sure what will happen if you throw an error inside array_map
 
What _could_ happen apart from the usual exception handling logic (that an exception has to be caught somewhere up the call stack or it results in a Fatal Error)? This is a genuine question, I have no idea what could be a problem here.

src/Cli/AbstractCli.php
```diff
@@ -912,4 +921,43 @@ public function prepareArgsManual(array $args): string
...
+
+		$ds = DIRECTORY_SEPARATOR;
+
+		foreach ($iterator as $item) {
+			$subPathName = $iterator->getSubPathname();
+			$destinationPath = rtrim($destination, $ds) . $ds . $subPathName;
+
+			if ($item->isDir()) {
+				if (!file_exists($destinationPath)) {
+					mkdir($destinationPath, 0777, true);
```
iobrado, line 956: Is there a specific reason this is set to `0777`?  Folders usually have permissions set to `0755`

iobrado (changes requested):
Left just one question. As soon as that's resolved, I'll approve 🙂

src/Cli/AbstractCli.php
dingo-d: I think by default the PHP sets it to 0777, we can change that to 0755, I think that's the default value that bash sets. Good catch 👍🏼

iobrado: We should definitely test how it behaves if set to 0755, just to be sure it doesn't break anything. Generally, having 0777 is a potential security risk since everyone has read, write and execute permissions.

goranalkovic-infinum (changes requested):
Left one suggestion and 2 missing thing comments 😄 

Awesome job!

src/Cli/AbstractCli.php
iruzevic: fixed

iobrado (approved):

goranalkovic-infinum (approved):

simun-ivanac (approved):

mbmjertan (approved):

tests/data/src/Blocks/components/button-false/manifest.json
```diff
@@ -0,0 +1,46 @@
+{
```
volfkarlo, line 1: Spaces instead of tabs in this whole file 😄

tests/Blocks/BlocksExampleTest.php
```diff
@@ -7,6 +7,8 @@
 use EightshiftBoilerplate\Blocks\BlocksExample;
 use EightshiftLibs\Exception\InvalidBlock;
 
+use EightshiftLibs\Helpers\Components;
```
volfkarlo, line 10: Extra new line above this one

tests/Blocks/BlockVariationCliTest.php
```diff
@@ -4,28 +4,32 @@
 
 use EightshiftLibs\Blocks\BlockVariationCli;
 
+use EightshiftLibs\Exception\InvalidBlock;
+
```
volfkarlo, line 8: extra new line above this one

tests/Blocks/BlockComponentCliTest.php
```diff
@@ -4,28 +4,33 @@
 
 use EightshiftLibs\Blocks\BlockComponentCli;
 
+use EightshiftLibs\Exception\InvalidBlock;
+
```
volfkarlo, line 8: extra new line above this one

tests/Blocks/BlockCliTest.php
```diff
@@ -4,13 +4,18 @@
 
 use EightshiftLibs\Blocks\BlockCli;
 
+use EightshiftLibs\Exception\InvalidBlock;
+
```
volfkarlo, line 8: extra new line above this one

src/Setup/UpdateCli.php
```diff
@@ -101,4 +101,116 @@ public function __invoke(array $args, array $assocArgs) // phpcs:ignore
 			exit("{$e->getCode()}: {$e->getMessage()}"); // phpcs:ignore Eightshift.Security.CustomEscapeOutput.OutputNotEscaped
 		}
 	}
+
+
```
volfkarlo, line 105: extra new line.

volfkarlo (commented):
Just suggestions for the next merge 😄

PR #265, opened 2022-03-18
Feature/linter

dingo-d:
# Description

Fixing linter issues - phpcs and phpstan are fixed, but tests seem to be failing. 

This needs to be addressed.

I think that adding the `\defined('ES_DEVELOP_MODE') ? \ES_DEVELOP_MODE : false` to the parameters caused issues.

TBH not sure why this was added in the first place, but all of those should be checked, so that we see which are optional, and which are not optional arguments, because I think in some cases this should be 

```
\defined('ES_DEVELOP_MODE') ? \ES_DEVELOP_MODE : true
```

### EDIT:

Fixed the failing tests except the Components one. For some reason the 

```php	
	private static function getManifestDirect(string $path): array
	{
		$sep = \DIRECTORY_SEPARATOR;
		$path = \trim($path, $sep);

		$manifest = "{$path}{$sep}manifest.json";

		if (!\file_exists($manifest)) {
			throw ComponentException::throwUnableToLocateComponent($manifest);
		}

		return \json_decode(\implode(' ', (array)\file($manifest)), true);
	}```

method throws an error that the manifest is missing, but the path is correct and exists when I stop debbug the path 🤷🏼‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a 
 - Option to override default block class prefix in the blocks global settings manifest.
 
 ### Fixed
-- Changed directory separator from `/` to `DIRECTORY_SEPARATOR` constant to be able to work cross-platform.
+- Changed directory separator from `/` to `\DIRECTORY_SEPARATOR` constant to be able to work cross-platform.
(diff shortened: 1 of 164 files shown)


iruzevic:
I will test this

iruzevic (approved):

PR #268, opened 2022-03-24
Fix for ACF is admin showing fields

volfkarlo:
# Description
- removed ACF is_admin early exit due to not serving ACF fields on the frontend

# Screenshots / Videos
is admin blocking acf fields
without is_adminacf

Note:
This was causing the user image not to be shown on frontend.
@goranalkovic-infinum, @iobrado

### DIFF
--- a/src/CustomMeta/AbstractAcfMeta.php
+++ b/src/CustomMeta/AbstractAcfMeta.php
@@ -29,10 +29,6 @@ abstract class AbstractAcfMeta implements ServiceInterface
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


dingo-d:
@volfkarlo but this class is only for registering ACF admin fields, it has nothing to do with the frontend part of the application 👀

volfkarlo:
We can jump on a call and try to resolve it. 
I've been on a call with @goranalkovic-infinum, @iobrado and @iruzevic about this problem. Only by removing this I managed to resolve the issue.

PR #272, opened 2022-03-25
Improve developer experience on Windows

mbmjertan:
# Description

This PR:
* replaces instances in which `wp boilerplate setup_theme` uses `/` instead of `DIRECTORY_SEPARATOR`
* stops prepending `DIRECTORY_SEPARATOR` to absolute paths on Windows
  * this caused `\C:\Users\...` paths on Windows, which were failing
  * the Windows check is based upon checking whether `\` is the directory separator, as we only care what's the filesystem structure like 
* stops quoting arguments when calling `wp boilerplate use_block` and `use_component` from `setup_theme`, as this these quotes for some reason aren't stripped on Windows
  * I can see that there were some similar bugs in WP-CLI regarding this, but WP-CLI doesn't officially support Windows so I didn't report this additionally
* includes changes from #270 

Using this, I was able to set up a theme (under Windows 11, using XAMPP) in the same way as I would on Unix systems.

### DIFF
--- a/src/Blocks/AbstractBlocksCli.php
+++ b/src/Blocks/AbstractBlocksCli.php
@@ -38,7 +38,8 @@ abstract class AbstractBlocksCli extends AbstractCli
 		$root = $this->getProjectRootPath();
 		$rootNode = $this->getFrontendLibsBlockPath();
 
-		$sourcePathFolder = "{$rootNode}/{$outputDir}/";
+		$ds = DIRECTORY_SEPARATOR;
+		$sourcePathFolder = "{$rootNode}{$ds}{$outputDir}{$ds}";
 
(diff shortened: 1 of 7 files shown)


gabriel-glo:
Looks good to me 👍 , but did you test this branch on both MacOS and Windows. Just so that we don't break one setup by fixing the other.

dingo-d (dismissed):

dingo-d:
Please rebase on the current develop branch and fix the failing issues in the actions 🙏🏼

dingo-d (approved):

gabriel-glo (approved):

dingo-d (approved):

gabriel-glo (approved):

PR #273, opened 2022-04-05
Feature/variables

iruzevic:
# Description

- updating css variables and inline logic to use our internal store for storing data.
- refactoring blocks abstract in full
- breaking component helper in multiple files

MISSING TESTS

Connected to https://github.com/infinum/eightshift-frontend-libs/pull/600


dingo-d (approved):
Looks good to me, I'd like to see the performance using blackfire for the store, to see if we store getters are called mulltiple times, and if so if we can cache them somehow (memoization, transient, object cache or something).

dingo-d (dismissed):

iobrado (approved):

dingo-d (dismissed):

dingo-d (approved):

PR #279, opened 2022-04-14
Add integration tests

dingo-d:
# Description

This PR adds integration tests to our libs 🎉🎉

What I've done is:

1. Pull the wordpress-develop to a `wp` subdirectory
2. Use a lib that adds a sqllite DB driver that should serve as a database for the WP (I need to test this)
3. Modify the `bootstrap.php` file so that unit tests work, but also integration tests are properly loaded
4. Update the `Pest.php` so that the correct base class is used in integration tests
5. Split unit and integration tests into their own respective folders so that I can group them

I need to run some more tests (creating users etc.) to see if there are any issues, but so far so good 🤞🏼

test1

test2


tests/bootstrap.php
```diff
@@ -0,0 +1,44 @@
...
+if (isset($GLOBALS['argv']) && isset($GLOBALS['argv'][1]) && strpos($GLOBALS['argv'][1], 'integration') !== false) {
+	if (!file_exists(dirname(__FILE__, 2) . '/wp/tests/phpunit/wp-tests-config.php')) {
+		// We need to set up core config details and test details
+		copy(dirname(__FILE__, 2) . '/wp/wp-tests-config-sample.php', dirname(__FILE__, 2) . '/wp/tests/phpunit/wp-tests-config.php');
+
+		// Change certain constants from the test's config file.
+		$testConfigPath = dirname(__FILE__, 2) . '/wp/tests/phpunit/wp-tests-config.php';
+		$testConfigContents = file_get_contents($testConfigPath);
+
+		$testConfigContents = str_replace("dirname( __FILE__ ) . '/src/'", "dirname(__FILE__, 3) . '/src/'", $testConfigContents);
```
mbmjertan, line 23: `DIRECTORY_SEPARATOR`

tests/bootstrap.php
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	// Modify the bootstrap so that Pest works.
+	$basicBootstrapPath = dirname(__FILE__, 2) . '/wp/tests/phpunit/includes/bootstrap.php';
+	$basicBootstrapContents = file_get_contents($basicBootstrapPath);
+
+	// We need to remove these because we already included them in Pest.php
+	$basicBootstrapContents = str_replace("require __DIR__ . '/phpunit-adapter-testcase.php';", '', $basicBootstrapContents);
+	$basicBootstrapContents = str_replace("require __DIR__ . '/abstract-testcase.php';", '', $basicBootstrapContents);
+	$basicBootstrapContents = str_replace("require __DIR__ . '/testcase.php';", '', $basicBootstrapContents);
```
mbmjertan, line 39: `DIRECTORY_SEPARATOR`

mbmjertan:
Wpizardry, really 😂

tests/bootstrap.php
dingo-d: This one comes bundled with WordPress integration suite, so I'm not sure I should be changing them. 

I definitely shouldn't use DS in string replace (because it's parsed as a string and I need to remove these (because I'm including them in Pest.php file manually 🙂

tests/bootstrap.php
dingo-d: I can add it, but there are other slashed files out of the box in here (from the core)

tests/Integration/Test.php
```diff
@@ -0,0 +1,31 @@
...
+
+    $this->server = $wp_rest_server = new \WP_REST_Server;
+    do_action( 'rest_api_init' );
+});
+
+/**
+ * Cleanup after tests.
+ */
+afterEach(function () {
+	global $wp_rest_server;
```
volfkarlo, line 19: not sure if missed indent.

volfkarlo (approved):

tests/bootstrap.php
mbmjertan: Oh, okay, wasn't aware of that context. I'd wager it's okay if the test suite doesn't run on Windows, but in that case, we should make it clear that's the case.

dingo-d:
Integration tests are not possible (at least not using Pest at the moment) for the libs.

The issue is that the bootstrap sets up the WordPress, then the test is executed which sets up the theme code, at which point the request already passed the WordPress registration process so those files are not recognized by the core.

Ideally, if we could use process isolation in Pest (still not possible), we'd just run one empty test that only sets up the code to be tested, then in a separate process check if this new setup code is now registered correctly.

Process isolation will (probably) come to Pest when PHPUnit v10 is released, but unfortunately for us, that means that we can only test this on PHP 8.1+, as PHPUnit 10 will have that requirement (and Pest as well).

PR #280, opened 2022-04-15
Move reusable blocks menu to AdminMenus and create CLI command for its generation

simun-ivanac:
# Description

Adding reusable blocks menu page is no longer positioned inside `Blocks.php`, but it's moved to its own class inside `AdminMenus`. Also, I created new CLI command to simplify its creation.

New command `wp boilerplate create_admin_reusable_blocks_menu` accepts next arguments:

- title _(Reusable Blocks)_
- menu_title _(Reusable Blocks)_
- capability _(edit_posts)_
- menu_icon _(dashicons-editor-table)_
- menu_position _(4)_

Regarding slug, it seems it's fixed `'edit.php?post_type=wp_block'` and you can't have multiple Reusable block menu pages. For that reason all arguments are optional.

If I overlooked something, feel free to point it out. 

And yes, I'll write tests :smile:


mbmjertan (dismissed):

src/AdminMenus/AbstractAdminReusableBlocksMenu.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
dingo-d, line 1: Why do you need an abstract class for this? Do you plan on having multiple reusable block menus?

Abstract classes in libs are what is known as a [template method design pattern](https://en.wikipedia.org/wiki/Template_method_pattern). If you don't need to create different instances of the same thing, you don't need an abstract class 🙂

dingo-d (changes requested):
Left one comment, otherwise good job 👍🏼

iobrado (dismissed):

iruzevic (changes requested):
Not 100% happy with this approach. I will review it in dept later

dingo-d:
> Not 100% happy with this approach. I will review it in depth later

What approach would you prefer?

iruzevic:
Im retracting my last statement, this is ok. Great yob

iruzevic (approved):

tests/AdminMenus/AdminReusableBlocksMenuCliTest.php
```diff
@@ -74,8 +99,8 @@
...
 	$key = 'shortdesc';
 
-	$this->assertIsArray($documentation);
-	$this->assertArrayHasKey($key, $documentation);
-	$this->assertArrayHasKey('synopsis', $documentation);
-	$this->assertSame('Generates reusable blocks admin menu class file.', $documentation[$key]);
+	expect($documentation)->toBeArray();
+	expect($documentation)->toHaveKey($key);
+	expect($documentation)->toHaveKey('synopsis');
+	expect($documentation[$key])->toBe('Generates reusable blocks admin menu class file.');
```
dingo-d, line 105: Expectations are chainable, meaning you can do:

```suggestion
	expect($documentation)->toBeArray()
		->toHaveKey($key)
		->toHaveKey('synopsis');
		
	expect($documentation[$key])->toBe('Generates reusable blocks admin menu class file.');
```

dingo-d (dismissed):
Just fix the phpstan issue 🙂 

Otherwise this looks great to me 👍🏼

dingo-d (approved):

iruzevic (approved):

PR #291, opened 2022-05-24
Feature/cli update

iruzevic:
# Description

- added examples, options, defaults on all commands.
- restructured all commands to follow the parent commands standard
- refactored a few commands to use template strings for replacement
- refactored tests
- refactored all comments and descriptions
- added new cli parent classes for better documentation
- added links to documentation and GH examples

# Screenshots / Videos

Screenshot 2022-05-23 at 09 16 25
s://user-images.githubusercontent.com/23283324/169968778-1c661f0e-0e83-43b4-aea9-3b75808d30a2.png)
Screenshot 2022-05-23 at 09 20 28

# Note 
I would like to refactor all cli commands with the new helper but it will take time so I will do this gradually


src/AdminMenus/AdminMenuCli.php
```diff
@@ -35,12 +58,29 @@ class AdminMenuCli extends AbstractCli
 	public function getDevelopArgs(array $args): array
 	{
 		return [
-			'title' => $args[1] ?? 'Admin Title',
-			'menu_title' => $args[2] ?? 'Admin Title',
-			'capability' => $args[3] ?? 'edit_posts',
-			'menu_slug' => $args[4] ?? 'admin_title',
-			'menu_icon' => $args[5] ?? 'dashicons-admin-generic',
-			'menu_position' => $args[6] ?? 100,
+			'title' => 'Test Title',
```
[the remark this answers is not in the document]
iruzevic: there is no need for them because I used some simple helpers to provide the data

src/Cli/AbstractCli.php
```diff
@@ -152,22 +156,85 @@ public function registerCommand(): void
...
+	 *
+	 * @param array<string, mixed> $array Array of args to check.
+	 * @param string $key Argument name to check.
+	 *
+	 * @return string
+	 */
+	public function getArg(array $array, string $key): string
+	{
+		return isset($array[$key]) ? (string) $array[$key] : $this->getDefaultArg($key);
+	}
```
dingo-d, line 208:
```suggestion
	/**
	 * Get one argument.
	 *
	 * Use assocArgs props and fallback to default if missing.
	 *
	 * @param array<string, string> $arguments Array of args to check.
	 * @param string $key Argument name to check.
	 *
	 * @return string
	 */
	public function getArg(array $arguments, string $key): string
	{
		return isset($arguments[$key]) ? (string) $arguments[$key] : $this->getDefaultArg($key);
	}
```

src/Cli/AbstractCli.php
```diff
@@ -152,22 +156,85 @@ public function registerCommand(): void
...
+
+	/**
+	 * Define default props for command.
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function getDefaultArgs(): array
+	{
+		return [];
+	}
```
dingo-d, line 194: Can you just make this an abstract method? Are there some CLI commands that don't have default arguments?

src/Cli/AbstractCli.php
```diff
@@ -152,22 +156,85 @@ public function registerCommand(): void
...
 		);
 	}
 
 	/**
-	 * Define default develop props
+	 * Define default develop props.
 	 *
 	 * @param string[] $args WPCLI eval-file arguments.
 	 *
 	 * @return array<string, mixed>
```
dingo-d, line 170: Can we be more specific about what this argument array contains? Can it really contain everything or just strings, numbers, booleans?

Because I'm unsure if the `$args[$i] ?? $value` check is valid on line 178 (if it's a non nullable entity this will never default to `$value`).

src/Cli/AbstractCli.php
```diff
@@ -152,22 +156,85 @@ public function registerCommand(): void
...
+	 */
+	public function getDefaultArg(string $key): string
+	{
+		$args = $this->getDefaultArgs();
+
+		if (!$args) {
+			return '';
+		}
+
+		return isset($args[$key]) ? (string) $args[$key] : '';
```
dingo-d, line 226: If the arguments are always strings, casting isn't necessary.

Are the argument values strings, or can they be something else?

src/Cli/AbstractCli.php
iruzevic: yep there are bunch of them that is why I left this like this

tests/WpCli/WpCliTest.php
```diff
@@ -49,21 +49,5 @@
...
-	$this->assertArrayHasKey($descKey, $documentation);
-	$this->assertArrayHasKey($synopsisKey, $documentation);
-	$this->assertIsArray($documentation[$synopsisKey]);
-	$this->assertSame('Generates custom WPCLI command in your project.', $documentation[$descKey]);
-
-	$this->assertSame('assoc', $documentation[$synopsisKey][0]['type']);
-	$this->assertSame('command_name', $documentation[$synopsisKey][0]['name']);
-	$this->assertSame('The name of cli command name. Example: command_name.', $documentation[$synopsisKey][0]['description']);
-	$this->assertSame(false, $documentation[$synopsisKey][0]['optional']);
+	expect($this->customCommand->getDoc())->toBeArray();
```
dingo-d, line 52: This is terse 😂 

No love for checking the output? 😢

tests/View/EscapedViewCliTest.php
```diff
@@ -42,20 +41,9 @@
 	$this->assertNotEmpty($generatedEscapedView);
 	$this->assertStringContainsString('class EscapedView extends AbstractEscapedView implements ServiceInterface', $generatedEscapedView);
 	$this->assertStringContainsString('register', $generatedEscapedView);
-		$this->assertStringNotContainsString('someRandomMethod', $generatedEscapedView);
-
+	$this->assertStringNotContainsString('someRandomMethod', $generatedEscapedView);
```
dingo-d, line 44:
```suggestion
	expect($generatedEscapedView)->not-> toContain('someRandomMethod');
```

tests/Cli/CliInitProjectTest.php
```diff
@@ -56,18 +56,12 @@
 	$commandName = $this->cliInitProject->getCommandName();
 
 	$this->assertIsString($commandName);
-	$this->assertSame(($this->cliInitProject)::COMMAND_NAME, $commandName);
+	$this->assertSame('project', $commandName);
```
dingo-d, line 59:
```suggestion
	expect($commandName)
		->toBeString()
		->toBe('project');
```

tests/CiExclude/CiExcludeCliTest.php
```diff
@@ -49,7 +49,7 @@
 	$ciexclude = $this->ciexclude;
 	$result = $ciexclude->getCommandName();
 
-	$this->assertStringContainsString('init_ci_exclude', $result);
+	$this->assertStringContainsString('ci_exclude', $result);
```
dingo-d, line 52:
```suggestion
	expect($result)->toContain('ci_exclude');
```

tests/Blocks/BlockWrapperCliTest.php
```diff
@@ -58,20 +58,11 @@
 	$wrapper = $this->wrapper;
 	$result = $wrapper->getCommandName();
 
-	$this->assertStringContainsString('use_wrapper', $result);
+	$this->assertStringContainsString('wrapper', $result);
```
dingo-d, line 61:
```suggestion
	expect($result)->toContain('wrapper');
```

tests/Blocks/BlockCliTest.php
```diff
@@ -50,20 +50,11 @@
 	$block = $this->block;
 	$result = $block->getCommandName();
 
-	$this->assertStringContainsString('use_block', $result);
+	$this->assertStringContainsString('block', $result);
```
dingo-d, line 53:
```suggestion
	expect($result)->toContain('block');
```

src/AdminMenus/AdminSubMenuCli.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+// phpcs:ignoreFile Generic.Files.LineLength.TooLong
```
dingo-d, line 3: You have effectively disabled this for all the other files, because you didn't re-enable the phpcs for this rule 🙈 

Either exclude files in the `phpcs.xml.dist`, or on the lines that are problematic please.

Ref: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-Ruleset#the-annotated-sample-file

src/Cli/ParentGroups/CliBlocks.php
```diff
@@ -0,0 +1,35 @@
...
+ *    # Copy block by name.
+ *    $ wp boilerplate blocks block --name='paragraph'
+ *
+ *    # Copy component by name.
+ *    $ wp boilerplate blocks component --name='heading'
+ *
+ *    # Copy wrapper.
+ *    $ wp boilerplate blocks wrapper
+ */
+class CliBlocks extends WP_CLI_Command
```
dingo-d, line 29: Would it be better to maybe have just one of these classes and constants inside instead of multiple classes that only hold constants inside?

Not sure what is gained with this granulation 🤷🏼‍♂️

src/Cli/TopLevelGroup/CliBoilerplate.php
```diff
@@ -0,0 +1,31 @@
...
+ * We use boilerplate for all examples commands but this command prefix deppends on you project setup.
+ * This prefix can be changed in the `(new Cli())->load('boilerplate');` function call.
+ * If you have changed the prefix to something else, please keep that in mind when reading the examples.
+ *
+ * ## RESOURCES
+ *
+ * Our documentation is located here:
+ * https://infinum.github.io/eightshift-docs/
+ */
+class CliBoilerplate extends WP_CLI_Command
```
dingo-d, line 29: Could the class constants in the parent groups be here?

dingo-d (changes requested):
I've reviewed 66/108 files 😅 

Most of the files are basically the same, so I don't think I need to go with a fine-tooth comb over it.

The comments I've left are more oriented towards the architecture/structure of the lib.

Great job on splitting the commands and adding examples 🙂  👍🏼

src/AdminMenus/AdminSubMenuCli.php
dingo-d: Correct! My bad, I replaced it with disable and enable.

Still, I'd add the pattern exclusion in the ruleset:

```xml
	<rule ref="Generic.Files.LineLength">
		<exclude-pattern>*/src/**/*Cli.php</exclude-pattern>
		<properties>
			<property name="lineLimit" value="180" />
			<property name="ignoreComments" value="true" />
		</properties>
	</rule>
```

than have the comment on top of every CLI file 😉

src/Build/BuildCli.php
```diff
@@ -84,6 +90,21 @@ public function getDoc(): array
...
 				],
 			],
+			'longdesc' => $this->prepareLongDesc("
+				## USAGE
+
+				Used for building you project to production ready version in one command. Generally used in GitHub Actions or any other tool for continuous integration. This file will be copied to your project root under the bin folder.
+
+				## EXAMPLES
+
+				# Create service class:
```
iobrado, line 100:
```suggestion
				# Copy file:
```

src/Build/BuildCli.php
```diff
@@ -84,6 +90,21 @@ public function getDoc(): array
...
+				$ wp boilerplate {$this->getCommandParentName()} {$this->getCommandName()}
+
+				## RESOURCES
+
+				Service class will be created from this example:
```
iobrado, line 105:
```suggestion
				File will be created from this example:
```

iobrado (changes requested):
Awesome work! I found some typos in descriptions and added suggestions to fix them.

src/AdminMenus/AdminSubMenuCli.php
iruzevic: awesome I like the rule approach :D

src/Cli/ParentGroups/CliBlocks.php
iruzevic: I have tried that first but the problem is how WP-CLI parses the doc comments on the commands and this is the only way I could make them work.

src/Cli/TopLevelGroup/CliBoilerplate.php
iruzevic: answered in the https://github.com/infinum/eightshift-libs/pull/291#discussion_r881417229

tests/WpCli/WpCliTest.php
iruzevic: I don't see the point checking the docs array, we have nothing to gain with that

src/Cli/AbstractCli.php
iruzevic: Updated int, string and bool

src/Cli/AbstractCli.php
iruzevic: agreed

iruzevic: i have tested this and it turns out you can get string,int or bool here but the output should always be string. So this type hint here is ok

iobrado (approved):

tests/WpCli/WpCliTest.php
dingo-d: Good for keeping track of changes 😉

dingo-d:
@iruzevic just check the `wp-json` for routes with this branch, I got some WP_CLI error message when I loaded it for testing.

src/Geolocation/GeolocationExample.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	/**
+	 * Get geolocation executable phar location.
+	 *
+	 * @return string
+	 */
+	public function getGeolocationPharLocation(): string
+	{
+		$path = __DIR__ . '/geoip2.phar';
```
dingo-d, line 46: Directory separator

src/Geolocation/GeolocationExample.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	/**
+	 * Get geolocation database location.
+	 *
+	 * @return string
+	 */
+	public function getGeolocationDbLocation(): string
+	{
+		$path = __DIR__ . '/GeoLite2-Country.mmdb';
```
dingo-d, line 63: Same as up

dingo-d (changes requested):
Geolocation is not tested. Since this is a bigger PR it would be good to focus on having tests 🙂

src/Geolocation/GeolocationExample.php
iruzevic: fixed

PR #295, opened 2022-05-24
Add captainHook package

### LINKED ISSUE #288: Remove package.json from the libs
### Describe your feature request

We currently have package.json in the backend libs that have 0 JS code. And the reason is just to run husky, which won't be run, unless you run `npm install` command.

This is totally unnecessary IMO, and if we want to have a husky like behavior in the backend libs we can use https://github.com/captainhookphp/captainhook instead.

### Describe the solution you'd like

Remove `package.json` completely, and optionally add https://github.com/captainhookphp/captainhook as a dev development package for pre-commit hook run.

We could then create a singular `test` command, that will run phpstan, phpcs and unit tests.

_No response_

### Environment info

_No response_

Yes

Yes

dingo-d:
Used instead of Husky package but for PHP.

Closes #288


dadadavorin (approved):

iruzevic (approved):
the package name is awesome :D

captainhook.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "pre-push": {
+        "enabled": false,
+        "actions": []
+    },
+    "pre-commit": {
+        "enabled": true,
+        "actions": [
+            {
+                "action": "\\CaptainHook\\App\\Hook\\PHP\\Action\\Linting",
```
mbmjertan, line 23: We can probably remove this action, as it only reports syntax errors, which PHPStan already handles as well.

PR #296, opened 2022-05-25
Geolocation service class

iruzevic:
# Description

- new service class for geolocation used to provide cookie set for user location based on the GeoLite2 Country database.


dadadavorin (approved):
Nice 👍  As I am aware - we are using the MaxMind database to compare the IP address to the country codes, and once downloaded to the project it stays the same.

Maybe it would be better to fetch the database from time to time, so it stays updated (WooCommerce does it like this if we turn on the geolocation in settings)?

iruzevic:
> Nice 👍 As I am aware - we are using the MaxMind database to compare the IP address to the country codes, and once downloaded to the project it stays the same.
> 
> Maybe it would be better to fetch the database from time to time, so it stays updated (WooCommerce does it like this if we turn on the geolocation in settings)?

how you handle db fetching is up to you on the project. I agree that this should be done but it is per project basis

src/Geolocation/GeolocationExample.php
```diff
@@ -0,0 +1,72 @@
...
+	 *
+	 * @return string
+	 */
+	public function getGeolocationPharLocation(): string
+	{
+		$path = __DIR__ . '/geoip2.phar';
+
+		if (!\file_exists($path)) {
+			// translators: %s will be replaced with the phar location.
+			return new Exception(\sprintf(\esc_html__('Missing Geolocation phar on this locaiton %s', 'eightshift-libs'), $path));
```
iobrado, line 50:
```suggestion
			return new Exception(\sprintf(\esc_html__('Missing Geolocation phar on this location %s', 'eightshift-libs'), $path));
```

src/Geolocation/GeolocationExample.php
```diff
@@ -0,0 +1,72 @@
...
+	 *
+	 * @return string
+	 */
+	public function getGeolocationDbLocation(): string
+	{
+		$path = __DIR__ . '/GeoLite2-Country.mmdb';
+
+		if (!\file_exists($path)) {
+			// translators: %s will be replaced with the database location.
+			return new Exception(\sprintf(\esc_html__('Missing Geolocation database on this locaiton %s', 'eightshift-libs'), $path));
```
iobrado, line 67:
```suggestion
			return new Exception(\sprintf(\esc_html__('Missing Geolocation database on this location %s', 'eightshift-libs'), $path));
```

iobrado (approved):
Found 2 typos 🙂

src/Geolocation/AbstractGeolocation.php
```diff
@@ -0,0 +1,230 @@
...
+	 *
+	 * @return array<mixed>
+	 */
+	public function getAdditionalCountries(): array
+	{
+		return [];
+	}
+
+	/**
+	 * Gets ip address by manual. General used for development and testing.
```
goranalkovic-infinum, line 107:
```suggestion
	 * Gets an IP address manually. Generally used for development and testing.
```

src/Geolocation/AbstractGeolocation.php
```diff
@@ -0,0 +1,230 @@
...
+				$output,
+				$adittionalLocaitons
+			);
+		}
+
+		return $output;
+	}
+
+	/**
+	 * Get geolocation 2 digit code proivided by the project.
```
goranalkovic-infinum, line 183:
```suggestion
	 * Gets the 2-digit location code provided by the project.
```

src/Geolocation/Readme.md
```diff
@@ -0,0 +1,23 @@
+# GeoLocation data
+
+We use [GeoIP2-php library](https://github.com/maxmind/GeoIP2-php) to get the correct countries based on the IP address of the user that we compare with the countries list that we get from the [DataHub list](https://datahub.io/core/country-list).
```
goranalkovic-infinum, line 3:
```suggestion
We use the [GeoIP2-php library](https://github.com/maxmind/GeoIP2-php) to get the user's location (country) based on their IP address. List of the countries is compared with the [DataHub list](https://datahub.io/core/country-list).
```

src/Geolocation/AbstractGeolocation.php
```diff
@@ -0,0 +1,230 @@
...
+				'value' => $code,
+				'group' => [
+					\strtoupper($code),
+				],
+			];
+		}
+
+		// Provide custom countries.
+		$adittionalLocaitons = $this->getAdditionalCountries();
+		if ($adittionalLocaitons) {
```
mbmjertan, line 172:
```suggestion
		$additionalLocations = $this->getAdditionalCountries();
		if ($additionalLocations) {
```

src/Geolocation/AbstractGeolocation.php
```diff
@@ -0,0 +1,230 @@
...
+			$output = \array_merge(
+				$output,
+				$adittionalLocaitons
```
mbmjertan, line 175:
```suggestion
				$additionalLocations
```

mbmjertan (changes requested):
Also some typos. 😀

volfkarlo (approved):

PR #297, opened 2022-05-26
WIP: adding loadmore route

iruzevic:
# Description

- added new dynamic data class to handle load more.

TODO:

- [x] Add class
- [x] Test taxonomy and other combinations
- [x] Add cli command
- [x] PHPStan
- [x] PHPCS
- [ ] Update tests

# Linked PR

This is a linked PR with frontend libs: https://github.com/infinum/eightshift-frontend-libs/pull/625


src/DynamicData/AbstractDynamicData.php
```diff
@@ -0,0 +1,203 @@
...
+		if (!isset($query['offset'])) {
+			$query['offset'] = \count($initialItems);
+		}
+
+		// Remove paged from query is set.
+		if (isset($query['paged'])) {
+			unset($query['paged']);
+		}
+
+		// Prevend accidental issues.
```
iobrado, line 196:
```suggestion
		// Prevent accidental issues.
```

iobrado (approved):

src/DynamicData/AbstractDynamicData.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @return string
+	 */
+	abstract public function getMappedDynamicData(string $type, array $response): string;
+
+	/**
+	 * Get Dynamic data ajax response callback.
+	 *
+	 * @return string
+	 */
+	public function getDynamicDataAjax(): string
```
dingo-d, line 36: Can we create a custom rest route for this instead? 🙏🏼

dingo-d (changes requested):
It would be better to use REST instead of AJAX. Easier to handle permissions and other security issues...

src/Enqueue/Theme/EnqueueThemeExample.php
```diff
@@ -59,4 +60,22 @@ public function getAssetsVersion(): string
...
 	}
+
+	/**
+	 * Get script localizations
+	 *
+	 * @return array<string, mixed>
+	 */
+	protected function getLocalizations(): array
+	{
+		$sep = \DIRECTORY_SEPARATOR;
```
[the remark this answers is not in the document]
iruzevic: hehehe nice :D I got PTSP of this separator, when I see slash it goes DIRECTORY_SEPARATOR :D

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+use EightshiftLibs\Rest\CallableRouteInterface;
+use EightshiftLibs\Rest\Routes\AbstractRoute;
+use WP_Error;
+use WP_Query;
+use WP_REST_Request;
+
+/**
+ * AbstractLoadMore class.
+ */
+abstract class AbstractLoadMore extends AbstractRoute implements CallableRouteInterface
```
dingo-d, line 22: What's the idea behind this abstract class?

If you want to have multiple load more routs, you'd extend it, but you'd still need to modify the response and the actual query logic for loading more posts/pages/entities?

I'm not sure I understand how you'd implement this.

Can you explain it?

iruzevic: The idea is to have one rest-route in your project where you map the response to your component and nothing else. In general you can have multiple load more instances of routes but in 99 % cases this will not be the case. 

In your project you will have only this src/Rest/Routes/LoadMore/LoadMoreRouteExample.php

iruzevic: Maybe this will help
https://github.com/infinum/eightshift-frontend-libs/tree/feature/block-featured-content/blocks/init/src/Blocks/components/load-more/docs

dingo-d: Ok, but in that case, this can just be a concrete `final` class with all the logic inside? 

I think that would be better.

We should limit abstract classes to templates, where it makes sense to use them. Always favoring composition over inheritance where possible.

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+			]
+		);
+
+		if (\is_wp_error($response)) {
+			return \wp_send_json_error(
+				[
+					'code' => $response->get_error_code(),
+					'msg' => $response->get_error_message(),
+				],
+				400
```
dingo-d, line 79: Can this code be extracted out the `WP_Error` object?

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+	private function getLoadMoreData(array $query, string $type, array $initialItems, string $urlLoadMoreAction, int $currentPage, array $config)
+	{
+		// Prepare parameters.
+		$perPageOverride = $config['perPageOverride'] ?? '';
+
+		// Bailout if query is missing.
+		if (!$query) {
+			return new WP_Error(
+				'error_missing_query',
+				\esc_html__('Your request is missing query param. Query param should be a valid WP_Query arguments.', 'eightshift-boilerplate')
```
dingo-d, line 117:
```suggestion
				\esc_html__('Your request is missing query parameters. Query parameters should be a valid WP_Query arguments array.', 'eightshift-boilerplate')
```

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+			);
+		}
+
+		// Bailout if type is missing.
+		if (!$type) {
+			return new WP_Error(
+				'error_missing_type',
+				\esc_html__('Your request is missing type param. This is used to filter whate response will the method return.', 'eightshift-boilerplate')
```
dingo-d, line 125:
```suggestion
				\esc_html__('Your request is missing type parameter. This is used to filter out what kind of response will the method return.', 'eightshift-boilerplate')
```

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+			);
+		}
+
+		// Bailout if initialItems is missing.
+		if (!$initialItems) {
+			return new WP_Error(
+				'error_missing_initial_items',
+				\esc_html__(
+					'Your request is missing initialItems param. This is used to determin number of items to load and what items to exclude from new requests.',
```
dingo-d, line 134:
```suggestion
					'Your request is missing initialItems parameter. This is used to determine the number of items to load and what items to exclude from new requests.',
```

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+					'eightshift-boilerplate'
+				)
+			);
+		}
+
+		// Prepare original query.
+		$originalQuery = $this->prepareOriginalQuery($query, $initialItems, $perPageOverride);
+
+		// Create new query to modify of url load.
```
dingo-d, line 143:
```suggestion
		// Create a new query to modify the load more URL.
```

Is this correct intention of the comment? I'm not sure I understood the original comment.

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+	 * @param array<string, mixed> $query Array of props for WP_Query.
+	 * @param array<string> $initialItems Array of initial items in dom.
+	 *
+	 * @return int
+	 */
+	private function getMaxCount(array $query, array $initialItems): int
+	{
+		$theQuery = new WP_Query($query);
+
+		\wp_reset_postdata();
```
dingo-d, line 201: Is this necessary? You have the `$newQuery` in the previous method. Can this be used to get the `$postsNumber`?

Just pass it as a parameter and you won't need to initialize the query again.

src/Rest/Routes/LoadMore/LoadMoreRouteExample.php
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * Get callback arguments array
+	 *
+	 * @return array Either an array of options for the endpoint, or an array of arrays for multiple methods.
+	 */
+	protected function getCallbackArguments(): array
+	{
+		return [
+			'methods' => static::CREATABLE,
```
dingo-d, line 60: Why is this `POST`?

You are retrieving the posts from the DB, this should be a `GET` (`READABLE`) endpoint.

dingo-d (changes requested):
Left a few comments. The REST API approach looks much better 😄

src/Rest/Routes/LoadMore/AbstractLoadMore.php
iruzevic: I think not https://developer.wordpress.org/reference/classes/wp_error/

src/Rest/Routes/LoadMore/AbstractLoadMore.php
iruzevic: i have updated the comment

src/Rest/Routes/LoadMore/AbstractLoadMore.php
```diff
@@ -0,0 +1,253 @@
...
+			);
+		}
+
+		// Prepare original query.
+		$originalQuery = $this->prepareOriginalQuery($query, $initialItems, $perPageOverride);
+
+		// Create new query to modify of url load.
+		$newQuery = $originalQuery;
+
+		// In case on url load update per page to new value.
```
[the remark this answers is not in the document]
iruzevic: i have updated the comment

src/Rest/Routes/LoadMore/AbstractLoadMore.php
iruzevic: done

dingo-d (commented):
From an architectural perspective looks good. We need tests for this.

Also, before merging, either do a squash merge so that we don't have a lot of useless commits, or rebase to clean up the commits a bit please 🙂

PR #298, opened 2022-06-03
adding partial helper

iruzevic:
# Description

- new partial helper that will load file from any component/block as a partial in a new folder used for small view components.

### DIFF
--- a/src/Exception/ComponentException.php
+++ b/src/Exception/ComponentException.php
@@ -58,4 +58,21 @@ final class ComponentException extends InvalidArgumentException implements Gener
 			)
 		);
 	}
+
+	/**
+	 * Throws exception if unable to locate partial.
+	 *
+	 * @param string $path Path.
+	 * @return static
+	 */
+	public static function throwUnableToLocatePartial(string $path): ComponentException
+	{
+		return new ComponentException(
+			\sprintf(
+			/* translators: %s is replaced with the path of the partial. */
+				\esc_html__('Unable to locate partial on this path: %s', 'eightshift-libs'),
+				$path
+			)
+		);
+	}
 }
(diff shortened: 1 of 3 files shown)


mbmjertan (approved):
😍

src/Exception/ComponentException.php
```diff
@@ -58,4 +58,21 @@ public static function throwUnableToLocateComponent(string $component): Componen
...
+	 * Throws exception if unable to locate partial.
+	 *
+	 * @param string $path Path.
+	 * @return static
+	 */
+	public static function throwUnableToLocatePartial(string $path): ComponentException
+	{
+		return new ComponentException(
+			\sprintf(
+			/* translators: %s is replaced with the path of the partial. */
```
dingo-d, line 72: Indentation seems off

src/Helpers/Components.php
```diff
@@ -151,6 +151,84 @@ public static function render(string $component, array $attributes = [], string
...
+		// Bailout if file is missing.
+		if (!\file_exists($path)) {
+			throw ComponentException::throwUnableToLocatePartial($path);
+		}
+
+		\ob_start();
+
+		require $path;
+
+		return \trim((string) \ob_get_clean());
```
dingo-d, line 229: It would be a good idea to check the ob level using 

```php
$level = ob_get_level();
```

before opening it, then before getting the contents to check if the `$level` is greater than another `ob_get_level` call, just to make sure somebody doesn't open an output buffer inside the view file and forgets to close it by accident.

iruzevic: I saw that we have mulitiple instances of buffer in the project. I will make an issue and implement the changes on all

PR #303, opened 2022-06-23
Add/fix install plugins method

dingo-d:
This PR adds the new plugin manage CLI method that should replace the update method. 

Should cover all the cases: install/update from wp.org (core), GitHub, and paid plugins. Also, has a `--delete-plugins` flag so that the only deletion can be made of the plugins which are not in the `setup.json`.

Update CLI method should probably be removed. 

> **Note** 
> The branch was branched off of the refactor PR #301. Once that is done and merged, I'll change the base of this PR to `develop`, and rebase it.

One thing I noticed but didn't know how to prevent is that the command requires the `--skip_existing=true` flag to be provided. Not sure why, but I'd like to remove this dependency if at all possible.


tests/Setup/PluginManageCliTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+	$wpCliMock
+		->shouldReceive('runcommand')
+		->withSomeOfArgs('plugin list --fields=name --format=json')
+		->andReturn([['name' => 'contact-form-7'], ['name'=>'wordpress-seo']]);
+
+	$wpCliMock
+		->shouldReceive('runcommand')
+		->withSomeOfArgs('plugin list --fields=name,version --format=json')
+		->andReturn([['name' => 'advanced-custom-fields-pro','version' => '5.10.2'],['name' => 'contact-form-7','version' => '5.1.8'],['name' => 'elementor-pro','version' => '3.7.0'],['name' => 'eightshift-forms','version' => '1.0.0'],['name' => 'query-monitor','version' => '3.6.7'],['name' => 'wp-rocket','version' => '3.11.2']]);
```
mbmjertan, line 41: I've had issues with using more than one `shouldRecieve` definition (only the first one would work, IIRC, as it would try to redefine the function twice). Does this work as expected?

dingo-d: Yup, works fine because it differs the two based on the `withSomeOfArgs` part 🙂

If it didn't work, the tests would fail 😄

mbmjertan (approved):

iruzevic (approved):

PR #306, opened 2022-06-28
Exctract handles to public methods

JerrySpecter:
# Description

Extract all handles to public methods so that they can be reused if needed.


iruzevic (approved):

dingo-d (commented):
Awesome, can you add tests for the new methods? Ping me if you need help 🙂

gabriel-glo (dismissed):

iobrado (approved):

PR #319, opened 2022-07-26
Return empty string in tests in getAssetsManifestItem

mbmjertan:
# Description

Fixes conditions for testing in `getAssetsManifestItem`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
           dependency-versions: highest
 
       - name: Run coverage generation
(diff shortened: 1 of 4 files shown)


iruzevic (dismissed):

dingo-d (approved):

dingo-d (changes requested):
Approved too quickly.

Please remove the `-q` flag from the GH Actions, so that we can see what actually fails (since we are no longer running just the unit tests without the coverage).

Also, can you check why the tests are failing?

iruzevic (approved):

dingo-d (approved):

PR #321, opened 2022-07-28
Feature/wrapper grid

iruzevic:
# Description

## Changed
- moving all "old" functions to legacy trait

## Added
- added new function to provide wrapper functionality to the core blocks.
- extracting common attributes to a separate function for reusability.
- new filter to example blocks file.
- new attribute that is passed via helpers `blockWrapClass `.


iobrado (dismissed):

src/Blocks/TraitBlocksLegacy.php
```diff
@@ -0,0 +1,78 @@
...
+
+namespace EightshiftLibs\Blocks;
+
+use EightshiftLibs\Helpers\Components;
+use WP_Post;
+
+/**
+ * Trait TraitBlocksLegacy
+ */
+trait TraitBlocksLegacy
```
dingo-d, line 19: Not sure why is this extracted to a trait. I don't see a benefit from having a trait here (in the Components class the traits were used to kinda clean it up - still not a perfect case for a trait, but we can live with it). Here I can just see the logic used in hooks extracted out - and this is definitely something we should avoid in the general code architecture decisions.

src/Blocks/BlocksExample.php
```diff
@@ -47,6 +47,7 @@ public function register(): void
 
 		// Filter block content.
 		\add_filter('render_block_data', [$this, 'filterBlocksContent'], 10, 2);
+		\add_filter('render_block_data', [$this, 'setNoneEightshiftBlocksWrapper'], 10, 2);
```
dingo-d, line 50: How come this logic is not in the `filterBlocksContent`? They are using the same hook 🤷🏼‍♂️

src/Blocks/AbstractBlocks.php
```diff
@@ -153,49 +166,49 @@ public function getAllBlocksList($allowedBlockTypes, WP_Block_Editor_Context $bl
...
 	 *
-	 * @throws InvalidBlock Throws error if blocks are missing.
+	 * @param array<string, mixed> $parsedBlock Array of block details.
+	 * @param array<string, mixed> $sourceBlock Array of block source details.
 	 *
-	 * @return void
+	 * @return array<string, mixed>
 	 */
-	public function registerBlocks(): void
+	public function setNoneEightshiftBlocksWrapper(array $parsedBlock, array $sourceBlock): array
```
dingo-d, line 188: The name of the method is odd. Maybe `setBlocksWrapperToNone`, or `disableWrapperForAllBlocks`?

src/Blocks/AbstractBlocks.php
```diff
@@ -153,49 +166,49 @@ public function getAllBlocksList($allowedBlockTypes, WP_Block_Editor_Context $bl
...
-		// Allow reusable block.
-		$allowedBlockTypes[] = 'core/block';
-		$allowedBlockTypes[] = 'core/template';
-
-		return $allowedBlockTypes;
 	}
 
 	/**
-	 * Method used to register all custom blocks with data fetched from blocks manifest.json.
+	 * Set none Eightshift block to use wrapper options.
```
dingo-d, line 181:
```suggestion
	 * Disable wrapper on all Eightshift blocks
```

dingo-d (changes requested):
Left some comments.

Also, we need to have tests for the new features added.

src/Blocks/AbstractBlocks.php
```diff
@@ -60,13 +65,21 @@ public function getBlocksDataFullRaw(): void
...
+				$namespace = $block['namespace'] ?? $settingsNamespace;
+
+				// Add common attributes to the store for later usage.
+				$commonAttributes = $this->getCommonAttributes($namespace, $block['blockName']);
+
 				$block['namespace'] = $namespace;
-				$block['blockFullName'] = "{$namespace}/{$block['blockName']}";
+
+				foreach ($commonAttributes as $attributeName => $attributeValue) {
+					$block[$attributeName] = $attributeValue['default'];
```
mbmjertan, line 81: What happens if an attribute does not have a default value?

PR #323, opened 2022-07-30
Fix plugins cli

### LINKED ISSUE #320: WP CLI command for plugin install/update is not working
### Describe your bug

Command parameters `[--skip_core=<skip_core>] [--skip_plugins=<skip_plugins>] [--skip_plugins_core=<skip_plugins_core>] …` are validated as boolean type, when they are strings type. 
There are multiple instances in `UpdateCli.php` that look like:

```
// Check if plugins key exists in config.
	if (!$skipPlugins) {
...
// Check if plugins core key exists in config.
	if (!$skipPluginsCore) {
...
```

Switching these conditionals to something like this:

```
// Check if plugins key exists in config.
	if ($skipPlugins === 'false') {

... 

if ($skipPluginsCore  === 'false') {
```

fixes the issue.

### Steps to Reproduce

1. Run `wp boilerplate run update` command
2. Nothing happens and I just get `Success: All commands are finished.` message

### Expected behavior

I expect to have WP updated, plugins from `setup.json` installed if not present (updated otherwise)

_No response_

### Environment info

_No response_

Yes

Yes

### Please confirm that your bug occurs with all plugins uninstalled and with the default WordPress theme active.

Yes

dingo-d:
# Description

Updated #303 PR. Closes #320

This PR adds the new plugin manage CLI method that should replace the update method.

Should cover all the cases: install/update from wp.org (core), GitHub, and paid plugins. Also, has a `--delete-plugins` flag so that the only deletion can be made of the plugins which are not in the `setup.json`.

I've had to refactor the wpcli mocks because if you alias the mock in the base class, then you cannot redefine them in individual tests, and if you define them all at once, they clash (the `runcommand` ones for some reason).

We'll have to check how properly to handle the dynamic properties that we are using across the tests in Pest, but this is something I'd like to tackle in a separate PR.


src/Setup/PluginManageCli.php
```diff
@@ -0,0 +1,548 @@
...
+	 *
+	 * @param string $name Identifier of the plugin from the setup.json file.
+	 *
+	 * @return string Plugin slug.
+	 */
+	private function getGithubPluginName(string $name): string
+	{
+		// If the plugin doesn't have a namespace, we're good, just return it.
+		if (\strpos($name, '/') === false) {
+			return $name;
```
mbmjertan, line 323: When can this be the case on Github? 👀

src/Setup/PluginManageCli.php
```diff
@@ -0,0 +1,548 @@
...
+		}
+
+		$splitName = \explode('/', $name);
+
+		return $splitName[\count($splitName) - 1];
```
mbmjertan, line 328: This and L345 will probably cause an issue? The repo owner is missing in the release URL that way?

E. g., this is the release URL for the latest Forms release: https://github.com/infinum/eightshift-forms/releases/download/1.4.1-alpha/release.zip - note that both `infinum` and `eightshift-forms` are constituent - this wouldn't be the case with this code if I'm not mistaken

src/Setup/PluginManageCli.php
dingo-d: Tbh not sure, this method is inside the deploy task so I took it from there. I guess we need some manual testing to see if this is necessary or not.

src/Setup/PluginManageCli.php
dingo-d: Again, not sure. I'd love it if somebody could test this in a real example (that is, using this branch when setting up a project, then filling up the `setup.json` file, and running the scripts) 🙏🏼 

All of the code here is in the deployment scripts (just modified to be a bit cleaner).

gabriel-glo (changes requested):
A few smaller things to fix. 
This brach should definitely be tested locally by at least two people before merge.

mbmjertan:
Tried it with GitHub and core plugins on a local project, found some issues:

## Wrong path to `setup.json` when running `plugin_manage`

Tries to look for `setup.json` in `wp-content/themes`, not the theme folder itself:

```
mariobornamjertan@Marios-MacBook-Pro ~/d/m/w/t/myproj> wp boilerplate run plugin_manage
Error: File missing on the path: /Users/mariobornamjertan/dev/multisitestandup/wp-content/themes/setup.json.
mariobornamjertan@Marios-MacBook-Pro ~/d/m/w/t/myproj [1]> pwd
/Users/mariobornamjertan/dev/multisitestandup/wp-content/themes/myproj
```

## Shell errors when running `plugin_manage`

```
mariobornamjertan@Marios-MacBook-Pro ~/d/m/w/t/myproj> wp boilerplate run plugin_manage
Remove plugins
Plugin akismet removed
Plugin hello removed
Plugin wp-rocket removed
Plugin advanced-cache.php removed
Success: Plugin query-monitor installed
Success: Plugin taxonomy-terms-order installed
Success: Plugin redirection installed
Success: Plugin wordfence installed
Success: Plugin wp-security-audit-log installed
Success: Plugin wp-crontrol installed
Success: Plugin wp-mail-smtp installed
Success: Plugin wordpress-seo installed
Downloading installation package from https://github.com/infinum/eightshift-forms/releases/download/1.3.0/release.zip...
Unpacking the package...
Installing the plugin...
Removing the old version of the plugin...
Plugin updated successfully.
Success: Installed 1 of 1 plugins.
sh: -c: line 0: syntax error near unexpected token `do'
sh: -c: line 0: `'/opt/homebrew/Cellar/php@7.4/7.4.30/bin/php' /usr/local/bin/wp  for f in ./wp-content/plugins/release*/; do rsync -avh --delete "$f" "./wp-content/plugins/eightshift-forms/" && rm -rf "$f"; done'
```

----

Regarding the Github plugin URL discussion, I misread that `shortName` was used instead of `name` in the URL. Sorry for the confusion 🙈

src/Setup/PluginManageCli.php
```diff
@@ -0,0 +1,548 @@
...
+					'description' => 'If you want to install only the plugins from github.org.',
+					'optional' => true,
+				],
+				[
+					'type' => 'flag',
+					'name' => 'install-paid',
+					'description' => <<<EOT
+					If you want to install only the paid plugins plugins.
+					You'll need an additional env.json file with premium plugin URLs.
+					EOT,
```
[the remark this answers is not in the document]
dingo-d: Since this isn't a `longdesc`, the only way to have a nice break line when checking the synopsis of the command is using [heredoc syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc).

dingo-d:
@mbmjertan thanks for the test, I'll try to reproduce this and fix the issues 👍🏼

src/Setup/PluginManageCli.php
dingo-d: Indeed this part is not needed at all.

dingo-d:
@mbmjertan Can you check if this works for you now? I've fixed the tests and parts that were failing, and also removed the unneeded rsync call

iruzevic (approved):

iruzevic (approved):

gabriel-glo (approved):

Issue #325, opened 2022-08-18
Remove translation functions from the CLI output commands


dingo-d:
### Describe your bug

When running some commands we get a nice output like this:

```bash
Success: Added block `accordion` at `wp-content/themes/wordpress-academy-course/src/Blocks/custom/`.

Dependency note:
We have found that this block has dependencies, please &#039;^&amp; run these commands also if you don&#039;t have it in your project:
wp boilerplate blocks use-component --name='accordion'
```

The issue is that this message is escaped:

```php
$this->cliLog(
	\sprintf(
		// translators: %s will be replaced with type of item.
		\esc_html__("We have found that this %s has dependencies, please '^& run these commands also if you don't have it in your project:", 'eightshift-libs'),
		$type
	)
);
```

Which will escape all sequences such as ampersand (`&`) or single quote (`'`). 

This is a bug, but this ticket is also a feature ticket to remove some escaping and all translating from the strings inside the CLI commands.

Some strings may need to be escaped (where we have any kind of input from the cli, we should sanitize that and escape maybe on output), but none of them should be translatable as we have no real way to translate the CLI commands (not sure if it's possible to do, and it definitely makes no sense to translate those).

### Steps to Reproduce

1. Setup project
2. Run init theme command
3. Run `wp boilerplate blocks use-block --name=accordion` command
4. See the escaped character

### Expected behavior

Nothing should be escaped.

### Please confirm that you have searched existing issues in this repo.

Yes

### Please confirm that you have searched in our documentation and couldn't find the answer.

Yes

### Please confirm that your bug occurs with all plugins uninstalled and with the default WordPress theme active.

Yes

dingo-d:
This has been merged to develop.

PR #326, opened 2022-08-18
Remove escaping from cliLog calls

dingo-d:
# Description

This PR fixes the issue in #325 

Luckily there were only two files where `esc_html__` was used. In other files, I've left it because those are mostly examples and abstract classes or exception classes.

### DIFF
--- a/src/Blocks/AbstractBlocksCli.php
+++ b/src/Blocks/AbstractBlocksCli.php
@@ -208,8 +208,7 @@ abstract class AbstractBlocksCli extends AbstractCli
 			$this->cliLog('Dependency note:', 'B');
 			$this->cliLog(
 				\sprintf(
-					// translators: %s will be replaced with type of item.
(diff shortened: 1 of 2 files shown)


goranalkovic-infinum (approved):

piqusy (approved):

mbmjertan:
Maybe it would be good to leave a `__` call in, so this message is translatable. I see no actual reason why someone would want to translate CLI logs (unless they're Microsoft 😄), but also no reason to not support that 🤷

gabriel-glo (approved):

dingo-d:
> Maybe it would be good to leave a `__` call in, so this message is translatable. I see no actual reason why someone would want to translate CLI logs (unless they're Microsoft 😄), but also no reason to not support that 🤷

It's not so much of no reason to support it. All of the other logs are not translatable, plus I'm honestly not sure if wp cli commands are translatable.

I see the CLI commands like status messages of REST API endpoints - you don't really translate them.

PR #331, opened 2022-08-24
Add standards:check Composer script

### LINKED ISSUE #330: Composer scripts are not inherited from dependencies
### Describe your bug

`composer run-script standards:check` is a non-existent script as it is not inherited from `infinum/eightshift-coding-standards`
That is not possible with Composer.

☣️ One more thing. This step does **not** fail in CI because of `continue-on-error: true`

### Steps to Reproduce

See CI step
https://github.com/infinum/eightshift-libs/runs/7931079011?check_suite_focus=true

### Expected behavior

PHPCS runs

_No response_

### Environment info

_No response_

Yes

No

### Please confirm that your bug occurs with all plugins uninstalled and with the default WordPress theme active.

No

szepeviktor:
# Description

Fixes #330

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Check coding standards using PHPCS
         continue-on-error: true
-        run: composer standards:check -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
+        run: composer test:standards -- --runtime-set testVersion ${{ matrix.php }}- --report-full --report-checkstyle=./phpcs-report.xml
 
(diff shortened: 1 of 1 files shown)


dingo-d:
This should actually be fixed in the workflow file. I've replaced the scripts but forgot to change it on GH Actions.

If you've noticed there is a `test:standards` script in the composer which checks the CS, but in the workflow the wrong one is referenced.

Can you fix the workflow instead to use the `test:standards` one?

szepeviktor:
> Can you fix the workflow instead to use the `test:standards` one?

@dingo-d Done 🍏

szepeviktor:
kép

dingo-d (approved):

PR #341, opened 2022-09-27
Adding geolocation corrections

iruzevic:
# Description

## Added
- New function for providing cookie expiration time.
- New file providing geolocation with WP-Rocket cache

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -82,8 +82,6 @@
 			"@test:standards",
 			"@test:types",
 			"@test:unit"
-		],
-		"analyze": "composer test:types",
-		"standards:check": "composer test:standards"
+		]
 	}
 }
--- a/composer.json
+++ b/composer.json
@@ -82,6 +82,8 @@
 			"@test:standards",
 			"@test:types",
 			"@test:unit"
-		]
+		],
+		"analyze": "composer test:types",
+		"standards:check": "composer test:standards"
 	}
 }
--- a/src/Geolocation/AbstractGeolocation.php
+++ b/src/Geolocation/AbstractGeolocation.php
@@ -59,6 +59,16 @@ abstract class AbstractGeolocation implements ServiceInterface
 	}
 
 	/**
+	 * Get geolocation expiration time.
+	 *
+	 * @return int
+	 */
+	public function getGeolocationExpiration(): int
+	{
+		return \time() + \DAY_IN_SECONDS;
+	}
+
+	/**
 	 * Set geolocation cookie.
 	 *
 	 * @return void
@@ -86,7 +96,7 @@ abstract class AbstractGeolocation implements ServiceInterface
 			$this->setCookie(
 				$cookieName,
 				$this->getGeolocation(),
-				\time() + \DAY_IN_SECONDS,
+				$this->getGeolocationExpiration(),
 				'/'
 			);
 		} catch (Exception $exception) {
--- a/src/Geolocation/Readme.md
+++ b/src/Geolocation/Readme.md
@@ -21,3 +21,83 @@ All releases are listed here: https://datahub.io/core/country-list
 
 Files used:
 * manifest.json
+
+## WP-ROCKET Plugin usage
+
+When geolocation is used in combination with WP-Rocket plugin it will not work by default because the cookie is set after the page is loaded and at that point it is too late.
+
+WP-Rocket plugin provides the options to fix this. By using this hook `rocket_advanced_cache_file` we are able to inject our custom function in the page generation process inside the `wp-content/advanced-cache.php` file. By modifying this file we can detect users geolocation from the IP and set the cookie manually before the page is loaded. This way WP-Rocket can detect that cookie and provide the necessary cache. Here is the process of setting it:
+
+1. Set dynamic cookie:
+
+**Filter:**
+```php
+\add_filter('rocket_cache_dynamic_cookies', [$this, 'dynamicCookiesList']);
+```
+
+**Callback:**
+```php
+/**
+ * List all dynamic cookies that will create new cached version.
+ *
+ * @param array<string, mixed> $items Items from the admin.
+ *
+ * @return array<int|string, mixed>
+ */
+public function dynamicCookiesList(array $items): array
+{
+	$items[] = 'esForms-country';
+
+	return $items;
+}
+```
+
+2. Set custom function to advanced-cached.php file.
+
+For `$esFormsPath` variable you must provide the absolute path to the `geolocationDetect.php` file in vendor.
+The provided example works if you have Eightshift-forms in your project.
+
+**Filter:**
+```php
+\add_filter('rocket_advanced_cache_file', [$this, 'addNginxAdvanceCacheRules']);
+```
+
+**Callback:**
+```php
+/**
+ * Add geolocation function in advance-cache.php config file on plugin activation used only with Nginx.
+ *
+ * @param string $content Original file output.
+ */
+public function addNginxAdvanceCacheRules( string $content ) : string {
+	$position = \strpos($content, '$rocket_config_class');
+
+	// This part is string on purpose.
+	$cors_function = '
+	$esFormsPath = ABSPATH . "wp-content/plugins/eightshift-forms/src/Geolocation/geolocationDetect.php";
+	if (file_exists($esFormsPath)) {
+		require_once $esFormsPath;
+		setLocationCookie(
+			ES_GEOLOCAITON_COOKIE_NAME,
+			\ABSPATH . "geolocation" . \DIRECTORY_SEPARATOR . "geoip.phar",
+			\ABSPATH . "geolocation" . \DIRECTORY_SEPARATOR . "geoip.mmdb",
+			ES_GEOLOCAITON_IP
+		);
+	};';
+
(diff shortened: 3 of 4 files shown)


src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+{
+	$items[] = 'esForms-country';
+
+	return $items;
+}
+```
+
+## Set custom function to advanced-cached.php file.
+
+By providing custom custom to advanced-cached.php file you will be able to detect geolocation and set cookies before the cached version is provided.
```
mbmjertan, line 59:
```suggestion
By providing a custom function in the advanced-cached.php file, you will be able to detect geolocation and set cookies before the cached version is provided.
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
 
 Files used:
 * manifest.json
+
+# WP-ROCKET Plugin usage
+
+When geolocation is used in combination with WP-Rocket plugin it will not work by default because the cookie is set after the page is loaded and at that point it is too late to provide content.
+
+WP-Rocket plugin provides the options/hooks to fix this. By using this hook `rocket_advanced_cache_file` we are able to inject our custom function in the page generation process inside the `wp-content/advanced-cache.php` file. By modifying this file we can detect users geolocation from the IP and set the cookie manually before the page is loaded. This way WP-Rocket can detect that cookie and provide the necessary cache. Here is the process of setting it:
```
piqusy, line 29:
```suggestion
WP-Rocket plugin provides options/hooks to fix this. By using hook `rocket_advanced_cache_file` we are able to inject our custom function in the page generation process inside the `wp-content/advanced-cache.php` file. By modifying this file we can detect users geolocation from the IP and set the cookie manually before the page is loaded. This way WP-Rocket can detect that cookie and provide the necessary cache. Below is the process of setting it:
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+
+## Set dynamic cookie:
+
+You must provide the list of cookies that will be used as a dynamic cookie to generate cached versions.
```
piqusy, line 33:
```suggestion
You must provide the list of cookies that will be used as dynamic cookies to generate cached versions.
```

src/Geolocation/Readme.md
piqusy, line 59:
```suggestion
By providing custom function to advanced-cached.php file you will be able to detect geolocation and set cookies before the cached version is provided.
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+		require_once $esFormsPath;
+	};';
+
+	return \substr_replace($content, $cors_function, $position, 0);
+}
+```
+
+## Disable `setLocationCookie` filter in your project.
+
+If you have installed Geolocation.php class in your project and you have this filter set, make sure you disable it, or provide the check to be disabled when WP-Rocket cache plugin is active.
```
piqusy, line 92:
```suggestion
If you have installed Geolocation.php class in your project and you have this filter set, make sure to disable it, or provide the check to be disabled when WP-Rocket cache plugin is active.
```

piqusy (changes requested):
Great work 🚀 Left some semantic & spelling suggestions.

piqusy (dismissed):

volfkarlo (approved):
Seems good! 😄

composer.json
```diff
@@ -82,6 +82,8 @@
 			"@test:standards",
 			"@test:types",
 			"@test:unit"
-		]
+		],
+		"analyze": "composer test:types",
+		"standards:check": "composer test:standards"
```
dingo-d, line 87: Why duplicate the commands you have under the `test` namespace?

src/Geolocation/AbstractGeolocation.php
```diff
@@ -86,7 +96,7 @@ public function setLocationCookie(): void
 			$this->setCookie(
 				$cookieName,
 				$this->getGeolocation(),
-				\time() + \DAY_IN_SECONDS,
+				$this->getGeolocationExpiration(),
```
dingo-d, line 99: Like the abstraction 👍🏼 more clear what the intention is.

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
 
 Files used:
 * manifest.json
+
+# WP-ROCKET Plugin usage
+
+When geolocation is used in combination with WP-Rocket plugin it will not work by default because the cookie is set after the page is loaded and at that point it is too late to provide content.
```
dingo-d, line 27:
```suggestion
When geolocation is used in combination with the WP-Rocket plugin it will not work by default because the cookie is set after the page is loaded and at that point, it is too late to provide alternative content.
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+
+WP-Rocket plugin provides options/hooks to fix this. By using hook `rocket_advanced_cache_file` we are able to inject our custom function in the page generation process inside the `wp-content/advanced-cache.php` file. By modifying this file we can detect users geolocation from the IP and set the cookie manually before the page is loaded. This way WP-Rocket can detect that cookie and provide the necessary cache. Below is the process of setting it:
```
dingo-d, line 29:
```suggestion
WP-Rocket plugin provides options/hooks to fix this. By using the `rocket_advanced_cache_file` hook we are able to inject our custom function into the page generation process inside the `wp-content/advanced-cache.php` file. By modifying this file we can detect the users' geolocation from the IP and set the cookie manually before the page is loaded. This way WP-Rocket can detect that cookie and provide the necessary cache. Below is the process of setting it:
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+{
+	$items[] = 'esForms-country';
+
+	return $items;
+}
+```
+
+## Set custom function to advanced-cached.php file.
+
+By providing a custom function in the advanced-cached.php file, you will be able to detect geolocation and set cookies before the cached version is provided.
```
dingo-d, line 59:
```suggestion
By providing a custom function in the `advanced-cached.php` file, you will be able to detect geolocation and set cookies before the cached version is provided.
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+
+For `$esFormsPath` variable you must provide the absolute path to the `geolocationDetect.php` file in vendor.
```
dingo-d, line 61:
```suggestion
For the `$esFormsPath` variable you must provide the absolute path to the `geolocationDetect.php` file located in the `vendor` folder.
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+\add_filter('rocket_advanced_cache_file', [$this, 'addNginxAdvanceCacheRules']);
+```
+
+**Callback:**
+```php
+/**
+ * Add geolocation function in advance-cache.php config file on plugin activation used only with Nginx.
+ *
+ * @param string $content Original file output.
+ */
```
dingo-d, line 75:
```suggestion
/**
 * Add geolocation function in the advance-cache.php config file on plugin activation
 *
 * Used only with Nginx web servers.
 *
 * @param string $content Original file output.
 */
```

src/Geolocation/Readme.md
```diff
@@ -21,3 +21,79 @@ All releases are listed here: https://datahub.io/core/country-list
...
+		require_once $esFormsPath;
+	};';
+
+	return \substr_replace($content, $cors_function, $position, 0);
+}
+```
+
+## Disable `setLocationCookie` filter in your project.
+
+If you have installed Geolocation.php class in your project and you have this filter set, make sure to disable it, or provide the check to be disabled when WP-Rocket cache plugin is active.
```
dingo-d, line 92:
```suggestion
If you have installed Geolocation.php class in your project and you have this filter set, make sure to disable it, or provide the check to be disabled when the WP-Rocket cache plugin is active.
```

dingo-d (changes requested):
Left some small remarks. Generally looks ok 👍🏼

composer.json
iruzevic: because we have that command in 30+ projects and libs, and only this project has it changed. I don't have the problem of having it like "test:" but I would like to retain the old command as well.

dingo-d: I was planning on changing this on other projects as well. I really don't like this change here, especially since it's a separate project on its own and the composer scripts are not shared between the projects.

We shouldn't revert the better change, instead we should improve other projects to have the better change...

dingo-d (approved):

PR #342, opened 2022-09-29
adding fixes for dependency in cli

iruzevic:
# Description

## Added
- WP-CLI will now look for an additional key when moving block/components. This key is used to list all inner block dependencies.

### DIFF
--- a/src/Blocks/AbstractBlocksCli.php
+++ b/src/Blocks/AbstractBlocksCli.php
@@ -201,7 +201,10 @@ abstract class AbstractBlocksCli extends AbstractCli
 		$manifest = Components::getManifestDirect($source);
 
 		// Component dependency.
-		$dependencies = $manifest['components'] ?? [];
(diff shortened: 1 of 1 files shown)


src/Blocks/UseStorybookCli.php
```diff
@@ -98,7 +94,24 @@ public function __invoke(array $args, array $assocArgs)
...
+	 * Common instalation steps.
+	 *
+	 * @return string
+	 */
+	private function getCommonAfterIntallInstructions(): string
+	{
+		return '
+			1. Run `npm install @eightshift/storybook --save-dev` command in the terminal to install the storybook package.
+			2. Open package.json.
+			3. Add to scrips: `"storybookBuild": "build-storybook -s public -o storybook"`
```
piqusy, line 112: Missing "t" in scripts. FOr some reason I cannot suggest

PR #350, opened 2022-12-22
Remove missing esc_html function from geolocationDetect

dadadavorin:
# Description

Removed escaping function `esc_html` which is not needed here and is causing errors.


dingo-d:
There are some CI failures, can you address those? What was the error/failure?

dadadavorin:
> There are some CI failures, can you address those? What was the error/failure?

@dingo-d the error was "Fatal error: Call to undefined function `esc_html`...".

As I can see those CI errors are not related to this PR, so I will sync before addressing this, to not interfere maybe with some other work currently being done for new release.

PR #361, opened 2023-05-22
Adds new command to WP CLI 

moonbyt3:
# Description

Adds new command to WP CLI that uses dummy block template to create new block by replacing all folders, file and their contents with block name provided as argument of WP CLI command.

## How to use new command:

Command for creating new **block** named hero:
```
wp boilerplate blocks rename --name='hero'  
```

Command for creating new **component** named hero:
```
wp boilerplate blocks rename --name='hero' --component='true'  
```


kancijan (commented):
Definitely, a feature I was looking for, but "rename" seems a bit misleading as I would expect it to rename an existing block or component in your project.

I would much rather like to have a `use-block` with the `--rename` option.
`wp boilerplate blocks use-block --name=example --rename="my-block"`

iruzevic:
I will check what I can do with this

PR #363, opened 2023-06-06
Reusable header/footer updates and CLI improvements

goranalkovic-infinum:
> **Note**
The commits make it look way scarier than it is (it was mostly just trying things out 😄)

Changes:
- added new CLI command for initializing reusable header/footer (settings page for picking blocks, and default blocks)
- added new helper to render nice alert "windows"
- revamped some CLI command outputs and error outputs

image
image
image


iobrado (approved):

piqusy (changes requested):
Nice work 👏🏼 
I left one suggestion 😉 

Other than that, I'm wondering why the header and footer need to be so closely coupled / created with the same command?

goranalkovic-infinum:
> Other than that, I'm wondering why the header and footer need to be so closely coupled / created with the same command?

They're this closely coupled because... well I was a bit lazy 😄

If you want to have both header and footer pickers on one screen, it would be a bit complicated to check whether you have none, one or both already added, and then try to patch files with the second picker.

That would also complicated things a bit on the Boilerplate side, where you would maybe need to check if it's even there.

Because every site will have some kind of a header and footer (and it's really ease to remove if for some reason it's not needed), it's combined into 1 command. (also, before full-auto theme setup I didn't want to have user run too many extra commands to set it up)

iobrado (dismissed):

iobrado (approved):

piqusy (approved):

iruzevic (approved):

PR #365, opened 2023-06-13
Fix deprecated PHPUnit method

szepeviktor:
# Description

New PHPUnit emits warnings.

# Screenshots / Videos

See CI output


dingo-d:
I wouldn't lock the versions of phpunit, let those be tied to PEST. Once Infinum's team decides to raise the minimum PHP version to 8.1, they can use pestv2 and higher phpunit version that comes as a dependency with it.

szepeviktor:
All right.

PR #371, opened 2023-08-02
added filter to getAllAllowedBlocksList

ghost:
# Description

We have a method to allow only blocks from the boilerplate, but sometimes we need to refine them, even more, eg. allow only paragraph block in Testimonial CPT. This filter allows us to do it in a service class instead of in the main Blocks class.

### DIFF
--- a/src/Blocks/AbstractBlocks.php
+++ b/src/Blocks/AbstractBlocks.php
@@ -131,7 +131,7 @@ abstract class AbstractBlocks implements ServiceInterface, RenderableBlockInterf
 		$allowedBlockTypes[] = 'core/block';
 		$allowedBlockTypes[] = 'core/template';
 
-		return \apply_filters('all_allowed_block_list', $allowedBlockTypes, $blockEditorContext);
+		return \apply_filters('es_allowed_blocks', $allowedBlockTypes, $blockEditorContext);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/Blocks/AbstractBlocks.php
```diff
@@ -131,7 +131,7 @@ function ($block) {
 		$allowedBlockTypes[] = 'core/block';
 		$allowedBlockTypes[] = 'core/template';
 
-		return $allowedBlockTypes;
+		return \apply_filters('all_allowed_block_list', $allowedBlockTypes, $blockEditorContext);
```
goranalkovic-infinum, line 134: Suggestion:
```suggestion
		return \apply_filters('es_allowed_blocks', $allowedBlockTypes, $blockEditorContext);
```
Just so it's not _that_ easy to confuse it with `allowed_block_types_all`

goranalkovic-infinum (dismissed):

goranalkovic-infinum (approved):

src/Blocks/AbstractBlocks.php
```diff
@@ -131,7 +131,7 @@ function ($block) {
 		$allowedBlockTypes[] = 'core/block';
 		$allowedBlockTypes[] = 'core/template';
 
-		return $allowedBlockTypes;
+		return \apply_filters('es_allowed_blocks', $allowedBlockTypes, $blockEditorContext);
```
iruzevic, line 134: this is not a good solution. the problem is that this filter is inside the libs. If you have plugin and theme in your project you have multiple instances of this filter. And if you assign something as a callback to this filter it will apply to plugin and theme and you don't want this.

If you need to change the default functionality from libs in your project you should instance a new function in you project and provide your logic.

I'm sorry but I will not approve this.

iruzevic (changes requested):

PR #373, opened 2023-09-22
Update libs with wpcs 3 changes [beta]

### LINKED ISSUE #315: PHP 8+ support
### Describe your feature request

We should start checking for PHP 8+ compatibility/issues in the libs.

### Describe the solution you'd like

1. Add GH Actions matrix checks, with allowed failures for PHP 8+ (since we are still not supporting them, it's ok that they fail)
2. Modify the phpcs config so that the supported version is upped to PHP8
3. Check if there are any forbidden keywords used as method arguments
4. Consolidate and agree on method argument naming (after PHP 8, any change in argument name will be a BC break!)
5. Update code to be written using PHP8 syntax with a fallback for PHP 7.4 version where possible
6. Test everything

_No response_

### Environment info

_No response_

Yes

Yes

dingo-d:
# Description

This PR will:

- Update coding standards to the latest beta version which should make the libs compatible with PHP 7.4, 8.0, 8.1, 8.2 (and the 8.3 tests seem to be passing)
- Fix the code so that the tests pass (the latest addition of adding private folders broke a lot of tests (95))
- Fix existing issues with phpcs and phpstan 
  - I've excluded some errors from phpcs reporting, as they are false positives, mainly the exception ones, as the exception messages are escaped in the exception classes
  - Also excluded file creation methods. The WP replacements are not ideal - [wp_mkdir_p()](https://developer.wordpress.org/reference/functions/wp_mkdir_p/) will create directories recursively but with 777 permission, and some other `WP_Filesystem_*` methods are just wrappers with filters added (which is not desired)
- Fix the deprecated test code:
  - assertions replaced with expectations to avoid breaking change of replacing `assertObjectHasAttribute` with `assertObjectHasProperty` in PHPUnit 9 (we could just create a custom assertion, but expectations are cleaner and don't suffer from this issue)
  - Added dynamically created properties to fix the [deprecation about dynamically created properties](https://www.php.net/manual/en/migration82.deprecated.php) in PHP 8.2+. The only deprecations about this come on versions where the composer was installed with the lowest versions of dependencies - `composer update --prefer-lowest --prefer-stable`, and that is due to Mockery mocks. On newer versions this issue [is fixed](https://github.com/mockery/mockery/issues/1209), but on older versions, it's not, and any on-the-fly property creation will cause this deprecation, which will fail on PHP9 (all deprecations will become fatal errors then).

Note that some tests do fail but those are _allowed failures_, and shouldn't matter as it's unlikely people would install the lowest stable packages anyhow.
Regarding the deprecation notices, I can, if you agree, just remove the `lowest` condition when installing the packages from the integrate.yml (would make the CI checks faster as well, as we would be removing extra checks).

Please review the code and test this code before tagging the v7 beta.

CC: @iruzevic @goranalkovic-infinum 

#### EDIT:

I've realized that the boilerplate will have to be updated with the beta release as well, as the coding standards would block installation (they are still on 1.6, which is locked to 7.4).

Closes #315

### DIFF
--- a/.github/workflows/integrate.yml
+++ b/.github/workflows/integrate.yml
@@ -63,8 +63,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version:
-          - ["7.4", "8.0", "8.1", "8.2"]
+        php-version: ["7.4", "8.0", "8.1", "8.2"]
         dependencies:
           - "lowest"
           - "highest"
--- a/.github/workflows/integrate.yml
+++ b/.github/workflows/integrate.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - name: "Checkout code"
-        uses: "actions/checkout@v3"
+        uses: "actions/checkout@v4"
 
       - name: "Check file permissions"
         run: |
@@ -40,7 +40,7 @@ jobs:
           coverage: "none"
 
       - name: "Checkout code"
-        uses: "actions/checkout@v3"
+        uses: "actions/checkout@v4"
 
       - name: "Find non-printable ASCII characters (box-drawing characters excluded)"
         run: |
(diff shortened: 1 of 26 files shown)


iruzevic:
I have created a new branch where we will test and update everything for PHP8, tnx for this major fixes

PR #381, opened 2023-10-26
WIP : PHP8

iruzevic:
⚠️ PHP8 preparation branch

# Description

- updating composer packages
- fixing stan and lint issues
- updating composer command names
- Denis fixes for standards
- adding missing type hinting
- Config changed definition of version and name
- Blocks removing "old" version functions
- new command to change version number
- fixed broken db import command with the wrong setup.json path
- adding better copy for checkAttr helper if key is missing.
- Every enqueue script now has `is<>Used` method that is by default true.

## Note

To install and test the latest version, use this command:
```sh
npx create-wp-project@latest --eightshiftLibsBranch="develop-php8" --eightshiftFrontendLibsBranch="develop-php8" --eightshiftBoilerplateBranch='develop-php8'
```

## Connected PRs
https://github.com/infinum/eightshift-boilerplate/pull/273
https://github.com/infinum/eightshift-frontend-libs/pull/761
https://github.com/infinum/eightshift-libs/pull/381


.github/workflows/integrate.yml
```diff
@@ -63,12 +63,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version:
-          - "7.4"
+        php-version: ["7.4", "8.0", "8.1", "8.2"]
         dependencies:
           - "lowest"
```
dingo-d, line 68: I think you can remove `lowest`, and then remove extra includes below with allowed failures. The tests should pass fine if that condition is removed (unless the code is legit broken ofc 😂 ).

composer.json
```diff
@@ -24,24 +24,24 @@
...
-		"infinum/eightshift-coding-standards": "^1.6",
+		"captainhook/captainhook": "^5.10",
+		"dealerdirect/phpcodesniffer-composer-installer": "^v1.0.0",
+		"infinum/eightshift-coding-standards": "2.0.0-beta",
 		"pestphp/pest": "^1.20",
 		"php-parallel-lint/php-parallel-lint": "^1.3",
-		"php-stubs/wordpress-stubs": "^5.9",
+		"php-stubs/wordpress-stubs": "^6.3",
 		"roave/security-advisories": "dev-master",
 		"szepeviktor/phpstan-wordpress": "^1.0.3",
```
dingo-d, line 43:
```suggestion
		"szepeviktor/phpstan-wordpress": "^1.3",
```

dingo-d (changes requested):
Left some minor comments.

The thing I'm more worried about is the change in the component rendering.
Several tests are failing (some of which are easy to fix), but for some, I'm not sure the behavior is correct, or at least this new feature has kinda complicated testing.

Because inside the tests you don't have WordPress, the test for the feature could be actually correct, but this new exception would make it seem like it's not correct.

For instance, the following test:

```php
test('Asserts that rendering a component works', function () {
	$results = Components::render('button', []);

	expect($results)
		->not->toBeEmpty()
		->toContain('Hello!');
});
```

This should be correct (confirming that rendering actually works), but this will now throw an exception: `You are not allowed to access paths outside of themes or plugins folder!`

So I think that either this should be fixed in test setup, to somehow 'trick' the code to think that during tests we are inside WordPress theme/plugin, or at the code level with a flag 🤷🏼‍♂️

.github/workflows/integrate.yml
iruzevic: removed

iruzevic:
> Left some minor comments.
> 
> The thing I'm more worried about is the change in the component rendering. Several tests are failing (some of which are easy to fix), but for some, I'm not sure the behavior is correct, or at least this new feature has kinda complicated testing.
> 
> Because inside the tests you don't have WordPress, the test for the feature could be actually correct, but this new exception would make it seem like it's not correct.
> 
> For instance, the following test:
> 
> ```
> test('Asserts that rendering a component works', function () {
> 	$results = Components::render('button', []);
> 
> 	expect($results)
> 		->not->toBeEmpty()
> 		->toContain('Hello!');
> });
> ```
> 
> This should be correct (confirming that rendering actually works), but this will now throw an exception: `You are not allowed to access paths outside of themes or plugins folder!`
> 
> So I think that either this should be fixed in test setup, to somehow 'trick' the code to think that during tests we are inside WordPress theme/plugin, or at the code level with a flag 🤷🏼‍♂️

I have fixed all the tests

.github/workflows/integrate.yml
iruzevic: not sure if I did it correctly, but all check are passing, can you please check

.github/workflows/integrate.yml
```diff
@@ -63,12 +63,11 @@ jobs:
...
       fail-fast: false
       matrix:
-        php-version:
-          - "7.4"
+        php-version: ["7.4", "8.0", "8.1", "8.2"]
         dependencies:
-          - "lowest"
           - "highest"
     runs-on: "ubuntu-latest"
+    continue-on-error: ${{ matrix.allowed_failure }}
```
dingo-d:
```suggestion
```

There is no `allowed_failure` key in the matrix so this can be removed.

dingo-d: You were faster than me 😄

iruzevic: It is all passing now

iruzevic:
I can't any errors in the tests, linters or stan. Also There are no issues in the debug log. Npx is working as expected :D I thing this setup is ready for PHP8.2. WooCommerce is also working :D

src/Db/DbImport.php
```diff
@@ -26,40 +26,44 @@ function dbImport(string $setupFile, array $args = [])
...
-			CliHelpers::cliError("Setup file is empty on this path: {$setupFile}.");
+			$errorClass::cliError("Setup file is empty on this path: {$setupFile}.");
 		}
 
 		// Check if urls key exists.
 		$urls = $data['urls'] ?? [];
 
 		if (empty($urls)) {
-			CliHelpers::cliError('Urls key is missing or empty.');
+			$errorClass::cliError('Urls key is missing or empty.');
```
dadadavorin, line 58: URLs

src/Db/DbImport.php
```diff
@@ -26,40 +26,44 @@ function dbImport(string $setupFile, array $args = [])
...
 		}
 
 		$fromHost = '';
 		$fromScheme = '';
 
 		// Die if from key is missing and not valid.
 		if (!isset($urls[$from]) || empty($urls[$from])) {
-			CliHelpers::cliError("{$from} key is missing or empty in urls.");
+			$errorClass::cliError("{$from} key is missing or empty in urls.");
```
dadadavorin, line 66: same

src/Db/DbImport.php
```diff
@@ -71,7 +75,7 @@ function dbImport(string $setupFile, array $args = [])
 
 		// Die if to key is missing and not valid.
 		if (!isset($urls[$to]) || empty($urls[$to])) {
-			CliHelpers::cliError("{$to} key is missing or empty in urls.");
+			$errorClass::cliError("{$to} key is missing or empty in urls.");
```
dadadavorin, line 78: also

dadadavorin (approved):

dadadavorin (approved):

piqusy (changes requested):
PHP8 🎆 
Nice work, left a few wording suggetions

tests/BaseTest.php
```diff
@@ -9,6 +9,7 @@
 use Mockery\MockInterface;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
+use WP_Theme;
```
dingo-d, line 12: Don't see this being used anywhere in the test 🤔

src/Blocks/AbstractBlocksCli.php
```diff
@@ -67,32 +67,42 @@ protected function moveItems(array $args, string $source, string $destination, s
...
+				}
+
+				// in the case of folders, we should also check the source folders.
+				if (\strpos($sourceFolder, $item) !== false) {
+					$itemExists = true;
+					break;
+				}
+			}
+
+
```
dingo-d, line 104:
```suggestion
```

tests/Unit/Exception/InvalidBlockTest.php
```diff
@@ -38,66 +44,80 @@
...
 
 	$missingNamespace = InvalidBlock::missingNamespaceException();
 
-	$this->assertIsObject($missingNamespace);
-	$this->assertObjectHasAttribute('message', $missingNamespace);
-	$this->assertSame('Global Blocks settings manifest.json is missing a key called namespace. This key prefixes all block names.', $missingNamespace->getMessage(), "Strings for message global settings manifest.json is missing a key called namespace do not match!");
+	expect($missingNamespace)->toBeObject()
+		->toBeInstanceOf(InvalidBlock::class)
+		->toHaveProperty('message')
+		->and('Global Blocks settings manifest.json is missing a key called namespace. This key prefixes all block names.')
```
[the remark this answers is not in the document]
dingo-d: Is this message changed in the CLI command? If not, this change would fail the test.

tests/Unit/Exception/InvalidBlockTest.php
```diff
@@ -38,66 +44,80 @@
...
 	$manifestPath = 'some/random/path';
 	$missingManifestPath = InvalidBlock::missingWrapperManifestException($manifestPath);
 
-	$this->assertIsObject($missingManifestPath);
-	$this->assertObjectHasAttribute('message', $missingManifestPath);
-	$this->assertSame("Wrapper blocks settings manifest.json is missing on this location: {$manifestPath}.", $missingManifestPath->getMessage(), "Strings for message if wrapper blocks settings manifest.json is missing do not match!");
+	expect($missingManifestPath)->toBeObject()
+		->toBeInstanceOf(InvalidBlock::class)
+		->toHaveProperty('message')
+		->and("Wrapper blocks settings manifest.json is missing on this location: {$manifestPath}.")
```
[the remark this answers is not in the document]
dingo-d: If this is changed in the CLI, then the message should be changed as well.

src/Misc/VersionCli.php
```diff
@@ -0,0 +1,122 @@
+<?php
```
dingo-d, line 1: This CLI command should be covered by tests.

src/Enqueue/Theme/AbstractEnqueueTheme.php
```diff
@@ -48,13 +48,31 @@ public function getThemeScriptHandle(): string
...
 	 * Register the Stylesheets for the front end of the theme.
 	 *
+	 * @param string $hook Hook name.
+	 *
 	 * @return void
 	 */
-	public function enqueueStyles(): void
+	public function enqueueStyles(string $hook): void
 	{
+		if (!$this->isEnqueueStylesUsed($hook)) {
```
dingo-d, line 72: I'm not sure I understand how this works, if the method will return `true` no matter what 👀

src/Enqueue/Blocks/AbstractEnqueueBlocks.php
```diff
@@ -172,13 +234,41 @@ public function enqueueBlockFrontendScript(): void
...
 	 * Enqueue blocks style for frontend only.
 	 *
+	 * @param string $hook Hook name.
+	 *
 	 * @return void
 	 */
-	public function enqueueBlockFrontendStyle(): void
+	public function enqueueBlockFrontendStyle(string $hook): void
 	{
+		if (!$this->isEnqueueBlockFrontendStyleUsed($hook)) {
```
dingo-d, line 268: How does this work, if all the method does is return `true`?

src/Enqueue/Admin/AbstractEnqueueAdmin.php
```diff
@@ -73,13 +80,41 @@ public function enqueueStyles(): void
...
 	 * Register the JavaScript for the admin area.
 	 *
+	 * @param string $hook Hook name.
+	 *
 	 * @return void
 	 */
-	public function enqueueScripts(): void
+	public function enqueueScripts(string $hook): void
 	{
+		if (!$this->isEnqueueScriptsUsed($hook)) {
```
dingo-d, line 114: Same as with the other helpers. This always returns `true`, not sure how this works 🤷🏼‍♂️

dingo-d (changes requested):
Left some minor comments and questions.

src/Db/DbImport.php
iruzevic: "urls" is a key in the setup.json so if i put it all caps someone might thing this is the key name. I would like it as is

tests/Unit/Exception/InvalidBlockTest.php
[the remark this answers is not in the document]
iruzevic: tnx but I would leave this as is for now. We have a ticket that will check all copy

tests/Unit/Exception/InvalidBlockTest.php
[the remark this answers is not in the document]
iruzevic: the same as before

tests/Unit/Cli/CliHelpersTest.php
```diff
@@ -5,5 +5,9 @@
 use EightshiftLibs\Cli\CliHelpers;
 
 test('cliError wrapper will return a WPCLI error', function () {
-	CliHelpers::cliError('Some random cli error happened');
+	$class = new class {
+		use CliHelpers;
+	};
+
+	$class::cliError('Some random cli error happened');
```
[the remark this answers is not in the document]
iruzevic: these are tests copy corrections are not necessary here

src/Blocks/AbstractBlocksCli.php
iruzevic: tnx

tests/BaseTest.php
iruzevic: removed

src/Misc/VersionCli.php
iruzevic: agreed will be in the next PR

src/Enqueue/Blocks/AbstractEnqueueBlocks.php
iruzevic: You can extend this function in your project and provide a condition when this script will be loaded. Hook param is passed from the original action. 

In general $hook is filled only in admin scripts not on the frontend. I have added this method on all scripts so you can potentially define every script where you would like to use it

piqusy (dismissed):

kancijan (dismissed):

src/Db/DbImport.php
```diff
@@ -26,40 +26,44 @@ function dbImport(string $setupFile, array $args = [])
 		$from = $args['from'] ?? '';
 		$to = $args['to'] ?? '';
 
+		$errorClass = new class () {
+			use CliHelpers;
+		};
```
mbmjertan, line 31: why is `CliHelpers` a trait to begin with? converting classes to traits just to instantiate anonymous classes implementing them to call trait methods smells a lot. if you want ping me and we can talk about what you want to achieve with this

tests/BaseTest.php
```diff
@@ -13,6 +13,47 @@
...
+	protected $regenerateWebPMediaCli;
+	protected $route;
+	protected $service;
+	protected $shortcode;
+	protected $themeEnqueue;
+	protected $webPMediaColumnCliMock;
+	protected $webPMediaColumnExampleMock;
+	protected $webPMediaColumnExampleMockColumns;
+	protected $wpRestServer;
+
```
mbmjertan, line 56: Just use the `AllowDynamicProperties` class attribute https://php.watch/versions/8.2/AllowDynamicProperties 
```suggestion
```

tests/BaseTest.php
```diff
@@ -13,6 +13,47 @@
 
 class BaseTest extends TestCase
```
mbmjertan, line 14:
```suggestion
#[AllowDynamicProperties]
class BaseTest extends TestCase
```

tests/Unit/Helpers/AttributesTraitTest.php
```diff
@@ -33,7 +33,7 @@
 	$attributes['buttonText'] = 'left';
 
 	Components::checkAttr('buttonAlign', $attributes, $manifest);
-})->throws(Exception::class, 'buttonAlign key does not exist in the button block manifest. Please check your implementation.');
+})->throws(Exception::class, 'buttonAlign key does not exist in the button block manifest. If you are using additional components, check if you used the correct block/component prefix in your attribute name. Please check your implementation.');
```
mbmjertan, line 36: better sentence order?
```suggestion
})->throws(Exception::class, 'buttonAlign key does not exist in the button block manifest. Please check your implementation. If you are using additional components, check if you used the correct block/component prefix in your attribute name.');
```

src/Cli/Cli.php
```diff
@@ -98,34 +99,35 @@ class Cli
 		AdminMenuCli::class,
```
mbmjertan, line 99: consider some simple autodiscovery for this, e.g. separate deeper namespaces and getting a list of classes in it. I understand that you rarely add any CLI commands but this seems like it's not the Eightshift way

mbmjertan (changes requested):

tests/BaseTest.php
dingo-d: When PHP 9 comes out, the attribute won't matter, because deprecations would turn to errors, so it's safer to just add these (granted the internals might just decide to just leave the attribute working even in PHP 9, but this is still a safer route).

I was thinking of just adding one `$mock` for all the mocked classes in the tests, but that would take me some time to check, cover and change, and this was faster 🤷🏼‍♂️

tests/Unit/Helpers/AttributesTraitTest.php
dingo-d: You'd have to also change the exception message in the exception itself otherwise the test would fail.

src/Cli/Cli.php
dingo-d: I was also wondering if this could be done in a smarter way, perhaps `glob`ing classes somehow, and just listing the `Cli` ones 🤔

src/Db/DbImport.php
iruzevic: I believe that while it may be possible to refactor this code for the sake of making it cleaner, it would not provide any significant value at this point. I will create an issue for it and revisit it later. Thank you for your comment.

iruzevic: https://github.com/infinum/eightshift-libs/issues/382

tests/BaseTest.php
iruzevic: thx for the suggestion but as these are tests I won't spend much time on refactoring this, and I agree with Denis

tests/Unit/Helpers/AttributesTraitTest.php
iruzevic: tnx, changed in both places

src/Cli/Cli.php
iruzevic: agree I will create a ticket for this https://github.com/infinum/eightshift-libs/issues/383

kancijan (approved):

goranalkovic-infinum (approved):

tests/BaseTest.php
dingo-d: Don't get me wrong, even tests should be clean, refactored, and taken care of, as they can acquire code cruft as time goes by (like this is a clear example of it). It's just that sometimes you need to do the 'dirty' fix to push the feature through, and fix it later 😄 

_We'll fix it in post_ 😂

piqusy (approved):

iobrado (approved):

tests/BaseTest.php
mbmjertan: The RFC somewhat implies that they will leave the attribute available in PHP 9, so I think that's a viable course of action. Either way, magic setters (__set) aren't deprecated and you could make a trait that allows setting dynamic properties. 

I understand your point of view and irrelevance of this in the grand scheme of things - but I believe we (Eightshift and friends) should strive to have a clean and understandable codebase.

iruzevic: agreed I will create a issue for this https://github.com/infinum/eightshift-libs/issues/384

PR #410, opened 2024-06-13
adding multiple fixes

iruzevic:
### Added
- Media support for svg and json.
- Optimization service class for better performance.

### Fixed
- Phpstan config fix for theme and plugin setup
- Project name variable will now be ucfirst on setup.
- Removed .git folder after the setup.

### DIFF
--- a/src/Media/AbstractMedia.php
+++ b/src/Media/AbstractMedia.php
@@ -148,22 +148,23 @@ abstract class AbstractMedia implements ServiceInterface
 	 */
 	public function validateSvgOnUpload($response)
 	{
-		if ($response['type'] === 'image/svg+xml' && \class_exists('SimpleXMLElement')) {
-			$path = $response['tmp_name'];
-
-			$svgContent = \file($path);
-			$svgContent = \implode(' ', $svgContent);
-
-			if (\file_exists($path)) {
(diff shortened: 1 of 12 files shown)


goranalkovic-infinum (dismissed):

src/Media/AbstractMedia.php
```diff
@@ -73,6 +78,128 @@ public function deleteWebPMedia(int $attachmentId): void
...
+			if (\file_exists($path)) {
+				if (!Helpers::isValidXml($svgContent)) {
+					return [
+						'size' => $response,
+						'name' => $response['name'],
+					];
+				}
+			}
+		}
+		return $response;
```
kancijan, line 166:
```suggestion
		if ($response['type'] !== 'image/svg+xml' && !\class_exists('SimpleXMLElement')) {
			return $response;
		}

		$path = $response['tmp_name'] ?? '';

		$svgContent = \file($path);
		$svgContent = \implode(' ', $svgContent);

		if (!\file_exists($path) || Helpers::isValidXml($svgContent)) {
			return $response;
		}

		return [
			'size' => $response,
			'name' => $response['name'],
		];
```

piqusy (dismissed):
LGTM

kancijan (changes requested):
Just one minor remark

PR #440, opened 2024-10-30
Cache Autowiring::buildServiceClasses into a transient

mbmjertan:
# Description

Using profiling we've gathered that we spend _a lot_ of time in `Autowiring::buildServiceClasses`. In particular, we spent 14% of the whole page generation time just running the method for `EightshiftForms`.

Here's a screenshot from an Excimer-grabbed profile opened up in Speedscope on my local machine:
Screenshot 2024-10-30 at 16 24 21

I added caching this into a transient using the project namespace as a unique key. This requires transients to be cleared on all deployments, and can cause breakage if `manuallyDefinedDependencies` is changed dynamically at runtime. This should probably be documented in eightshift-docs.

I should probably add a way to toggle this off. Do you prefer `WP_ENVIRONMENT_TYPE`, a special constant or both?

After this change, it's no longer noticeable in the profile and the whole `registerServices` method now takes less time than just the `buildServiceClasses` method used to:
Screenshot 2024-10-30 at 16 31 00

I will open up additional PRs for optimization opportunities if I notice them.

### DIFF
--- a/src/Cache/AbstractManifestCache.php
+++ b/src/Cache/AbstractManifestCache.php
@@ -159,13 +159,18 @@ abstract class AbstractManifestCache implements ManifestCacheInterface
 
 	/**
 	 * Get an (abstract) cache item.
+	 *
 	 * @param string $key Cache array key.
 	 * @param string $cacheType Cache type to get.
 	 * @param callable|null $fallback Fallback function for getting a result when not set. Optional.
 	 * @param bool $isJson Determines whether we should try to decode the cached value as JSON.
-	 * @return array|mixed[] An array of cached values.
+	 *
+	 * @throws InvalidManifest If cache item is missing.
+	 *
+	 * @return array<string, mixed> An array of cached values.
 	 */
-	public function getCacheTopItem(string $key, string $cacheType = self::TYPE_BLOCKS, callable $fallback = null, bool $isJson = true): array {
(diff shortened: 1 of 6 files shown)


dingo-d (dismissed):

iruzevic:
Thank you for the implementation; it’s a great idea! However, I wouldn’t make it a requirement to clear transients each time you deploy a manual thing. We already have manifest caching in the transients, along with cache busting based on the project version number and other factors. Since the Cache class is loaded before the Main class, I believe you could utilize the Cache class to get and set your transients effectively.

iruzevic (changes requested):
I left a comment about cache busting

mbmjertan:
Awesome, I wasn't familiar with how that class was implemented! ❤️ 
I will take a look and familiarize myself with the implementation and try to utilize that.

iruzevic:
@mbmjertan and progress here, you need help?

mbmjertan:
Hi, I didn't have the time to tackle this, hopefully I will today

mbmjertan:
Refactored to use AbstractManifestCache and added basic support for caching custom objects in AbstractManifestCache (but maybe a generic concrete caching implementation in Libs would make sense in the future). Please test this before merging.

mbmjertan:
We deployed this onto staging with no issues

PR #443, opened 2024-11-20
Paths not resolved correctly on Windows

ashthomas92:
On Windows, no paths are resolved from the manifest, leading to a InvalidManifest exception.

# Description

The `Helpers::joinPaths()` method prepends the directory separator onto the generated base file path. On Windows this causes the paths not to be resolved, leading to it being unable to read any files from `manifest.json`. This fix checks to see if the base path has a disk indicator prefixing the path, and only prepends the separator when it doesn't.

### DIFF
--- a/src/Helpers/Helpers.php
+++ b/src/Helpers/Helpers.php
@@ -461,7 +461,11 @@ class Helpers
 		);
 
 		$path = \implode($sep, $paths);
-		$path = "{$sep}{$path}";
+
+		// only add the separator if there is not a disk indicator - enables this to work in windows
+		if (!\str_contains(\mb_strstr($path, $sep, true) ?? '', ':')) {
+			$path = "{$sep}{$path}";
+		}
 
 		if (!\pathinfo($path, \PATHINFO_EXTENSION)) {
 			$path = "{$path}{$sep}";


mbmjertan:
Glad to see Eightshift usage and contributions.

I think this method is unreliable, especially for extending Eightshift, and should be refactored entirely. I am willing to do this myself.

This isn't at all a critique of your PR, @ashtomas92, and I am very grateful to see contributions.

- regarding your PR, filenames on Linux can contain `:`, so this check can break paths on Linux
- regarding the original implementation, why would all calls to this method need to include `ABSPATH` to be correct? can this not be a flag?

why couldn't I just do this:
```php
file_get_contents(Helpers::joinPaths(['Blocks', 'components', 'heading', 'manifest.json']
```
- we should probably also do some test coverage for Windows-particular filepath construction, given how often we've ran into issues with that

Once again, very happy to see contributions. I am hoping a fix will get merged soon.

hexydec:
If the original path is captured from a known location using `__DIR__`, the base path should be easy to work out without de/reconstructing it, the slash at the start can then just remain if it was already present, same for the drive indicator as per Windows.

ashthomas92:
Hi @iruzevic, just wondering if there's any been any movement on this?

iruzevic:
this will be totaly refactored in the release 10 so I will close this

PR #446, opened 2025-01-07
adding optimizations for webP check

iruzevic:
### Updated

- Updated `isWebPMediaUsed` method to use static variable so the check is done only once.

### DIFF
--- a/src/Helpers/MediaTrait.php
+++ b/src/Helpers/MediaTrait.php
@@ -59,7 +59,13 @@ trait MediaTrait
 	 */
 	public static function isWebPMediaUsed(): bool
 	{
-		return (bool) \get_option(UseWebPMediaCli::USE_WEBP_MEDIA_OPTION_NAME, true);
+		static $isWebPMediaUsed = null;
+
+		if ($isWebPMediaUsed === null) {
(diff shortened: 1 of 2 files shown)


iobrado (dismissed):

goranalkovic-infinum (approved):

src/Helpers/MediaTrait.php
```diff
@@ -59,7 +59,13 @@ public static function getWebPMedia(string $path, array $allowd = AbstractMedia:
 	 */
 	public static function isWebPMediaUsed(): bool
 	{
-		return (bool) \get_option(UseWebPMediaCli::USE_WEBP_MEDIA_OPTION_NAME, true);
+		static $isWebPMediaUsed = null;
+
+		if ($isWebPMediaUsed === null) {
+			$isWebPMediaUsed = (bool) \get_option(UseWebPMediaCli::USE_WEBP_MEDIA_OPTION_NAME, true);
```
goranalkovic-infinum, line 65: Isn't this condition always true?

goranalkovic-infinum (changes requested):

src/Helpers/MediaTrait.php
iruzevic: you are right it should be false. Fixed

iobrado (approved):

PR #452, opened 2025-02-20
Cache refactor

iruzevic:
# Description

- Restructuring caching system for manifest.
- Updating helpers for blocks.
- Fixing issues with linters.
- Simplifying blocks registration.
- Caching service classes to file like DI container
- refactoring render method to pass manifest and globalManifest variable to all content
- refactoring getProjectPaths method for simpler usage
- changing `PROJECT_RENDER_ALLOWED_NAMES` variables.
- deprecating `getManifestByDir` method
- removing all cli for blocks
- removing deprecated `renderPartial`, `getManifest` methods.


src/Helpers/CacheTrait.php
```diff
@@ -0,0 +1,232 @@
...
+	 */
+	public static function setCache(): void
+	{
+		if (self::$cache) {
+			return;
+		}
+
+		$output = [];
+
+		foreach (self::$cacheBuilder as $type => $value) {
```
piqusy, line 84: Why not `foreach (array_keys(self::$cacheBuilder) as $type) {` since $value is not used

piqusy (changes requested):
Nice work 👏🏼 

Left one suggestion

goranalkovic-infinum (approved):

piqusy (approved):

src/Db/DbImport.php
```diff
@@ -83,43 +83,41 @@ function dbImport(string $setupFile, array $args = [])
...
+		WP_CLI::runcommand("db import {$fileName}");
+		WP_CLI::log('Database import done.');
+		WP_CLI::log('--------------------------------------------------');
+
+		// Search and replace url host.
+		WP_CLI::runcommand("search-replace {$fromHost} {$toHost} --url={$fromHost} --all-tables --network");
+		WP_CLI::log('Database search replace for host successfully finished.');
+		WP_CLI::log('--------------------------------------------------');
+
+		// Search and replace url scheme.
```
piqusy, line 107:
```suggestion
		// Search and replace URL scheme.
```

src/Helpers/Helpers.php
```diff
@@ -84,15 +88,14 @@ class Helpers
 	 * @var array<int, string>
 	 */
 	private const PROJECT_RENDER_ALLOWED_NAMES = [
+		'src',
+		'blocksRoot',
```
piqusy, line 92: Do we need this here `blocksRoot`?

piqusy (changes requested):
Nice work 👏🏼 

Left a suggestion and a question

src/Main/Autowiring.php
```diff
@@ -79,17 +53,6 @@ public function setManifestCache(AbstractManifestCache $manifestCache = null): v
...
-			if (isset($this->manifestCache)) {
-				$serviceClasses = $this->manifestCache->getCacheTopItem('serviceClasses', AbstractManifestCache::TYPE_SERVICES, isJson: false);
-				if ($serviceClasses) {
-					return $serviceClasses;
-				}
-			}
-		} catch (InvalidManifest) {
-			// ignored, cache is simply not set.
-		}
-
```
mbmjertan, line 92: So service class caching will be entirely removed?

iruzevic: no, ti will be stored as a file cache like DI container

src/Main/AbstractMain.php
```diff
@@ -270,4 +256,56 @@ protected function getServiceClasses(): array
...
+			\file_put_contents( // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents
+				$cacheFile,
+				\wp_json_encode($services)
+			);
+
+			return $services;
+		}
+
+		return $services;
+	}
```
mbmjertan, line 293: Ah, I see, it's now cached to a file. But where is this set?

iruzevic: by default under `vendor-prefixed/infinum/eightshift-libs/src/Main/Cache` but you can use global contant to change the cache folder location if you have any issues with folder write permissions.

`define('EIGHTSHIFT_DI_CACHE_FOLDER', ABSPATH . 'eightshift-di-cache');` - this will store the cache in the project root

mbmjertan: ok I will have to read this in more depth to understand how it works, thx

PR #471, opened 2025-09-11
Release 11.0.0

iruzevic:
### Removed

- `WebPMediaColumnCli`, `WebPMediaColumnExample`, `MediaWebPTrait`, `UseWebPMediaCli` as it is not used anymore.

### Changed

- WebP helpers are now using totally different approach with different hooks and methods. We are no longer creating a new image on upload and using the helper method to detect the extension but we are creating a new image on upload and using using native WordPress functions.

### Updated

- composer packages.


dadadavorin (approved):
Looks good, nice work!

goranalkovic-infinum (approved):

piqusy (approved):

PR #488, opened 2026-04-24
Feature/type annotations

iobrado:
# Description

Replaces generic \`array<mixed>\` PHPDoc annotations with accurate inline types across 19 files in \`src/\`. Changes fall into a few categories:

- **String-keyed maps** (\`array<mixed>\` → \`array<string, mixed>\`): WP register args, manifest data, attributes, settings, admin/menu args, REST response arrays, media response arrays.
- **Typed list shapes** (\`array<mixed>\` → \`list<array{...}>\`): geolocation countries, block patterns — now carry explicit shape types with \`label\`, \`value\`, and \`group\` keys.
- **Nested cache structures**: \`$cache\` and \`$cacheBuilder\` narrowed to \`array<string, array<string, mixed>>\` and \`array<string, array<string, array<string, mixed>>>\` to reflect the three-level key hierarchy.
- **CSS variable internals**: \`$data\` params narrowed to \`array<int, array<string, mixed>>\`; \`setBreakpointResponsiveVariables\` return narrowed to \`list<array<string, mixed>>\`; \`variablesInner\` return narrowed to \`list<string>\`.
- **Union types instead of \`mixed\`**: \`$input\` in \`unifyClasses\` → \`string|array<string>\`; \`DeprecatedTrait\` API output returns → \`array<int|string, mixed>\`.
- **Precise shape annotations**: CLI \`$options\` → \`array{quality: int}\`; walker \`$children_elements\` → \`array<int|string, mixed>\`.

PHPStan passes cleanly after all changes.

## Why some `mixed` types remain

Four legitimate cases where `mixed` cannot be narrowed:

**1. `checkAttr` return type (`AttributesTrait:31`)**
Block attributes are genuinely polymorphic — a single attribute can be a `string`, `int`, `bool`, `array`, or `null` depending on what's declared in the manifest. PHPStan has no way to statically infer which, and callers handle all these cases. Narrowing to a union would be a lie; `mixed` is correct here.

**2. REST API return types (`CallableRouteInterface`, `CallableFieldInterface`, `RouteExample`)**
WordPress's `register_rest_route()` / `register_rest_field()` callbacks can return `WP_REST_Response|WP_Error|mixed` — this is the WP contract. Changing these would either misrepresent the interface or break PHPStan on the WP side.

**3. `ComponentException::$variable`**
The exception is designed to receive *any* invalid variable for diagnostic purposes. The variable that caused the type mismatch could be literally anything, so `mixed` is semantically correct.

**4. `CssVariablesTrait::variablesInner` `$attributeValue` param**
An attribute value piped into CSS variable resolution can be a `string`, `bool`, `int`, or `array`. Without a sealed union defined elsewhere in the system, `mixed` is the honest annotation — narrowing it would require changes to the attribute contract well beyond the scope of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a CHANGELOG](https://keepachangelog.com/).
 
+## [12.3.4]
+
+### Changed
+
+- Narrowed generic `array<mixed>` PHPDoc annotations to accurate inline types across 19 files — string-keyed maps, typed list shapes, nested cache structures, CSS variable internals, union types, and precise shape annotations.
+
 ## [12.3.3]
 
 ### Fixed
@@ -1130,6 +1136,7 @@ Init setup
 - Gutenberg Blocks Registration.
 - Assets Manifest data.
 
+[12.3.4]: https://github.com/infinum/eightshift-libs/compare/12.3.3...12.3.4
 [12.3.3]: https://github.com/infinum/eightshift-libs/compare/12.3.2...12.3.3
 [12.3.2]: https://github.com/infinum/eightshift-libs/compare/12.3.1...12.3.2
 [12.3.1]: https://github.com/infinum/eightshift-libs/compare/12.3.0...12.3.1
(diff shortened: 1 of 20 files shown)


iruzevic (approved):
I have tested this on infinum web and there weren't any  issues with stand or cs

goranalkovic-infinum (approved):

